PR #16, opened 2022-10-26
Use environment variables for `.condarc` values

ajschmidt8:
This PR updates the `.condarc` file to use values from environment variables. The environment variables serve as the single source of truth for these config values and can be easily referenced in downstream scripts that run inside our CI containers.


ajschmidt8:
xref: https://github.com/rapidsai/gha-tools/pull/22

ajschmidt8:
`gettext` is needed to provide the `envsubst` command

Ethyling (approved):

PR #23, opened 2022-11-23
Pin `rapids-dependency-file-generator` to major version

ajschmidt8:
With https://github.com/rapidsai/dependency-file-generator/issues/12 closed and semantic versioning enabled for `rapids-dependency-file-generator`, we can now pin to the latest major version in our CI images to prevent any releases with breaking changes from breaking our CI images.

v1.0.0 release - https://github.com/rapidsai/dependency-file-generator/releases/tag/v1.0.0

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -87,7 +87,7 @@ RUN cat /tmp/condarc.tmpl | envsubst | tee /opt/conda/.condarc; \
 RUN /opt/conda/bin/git config --system --add safe.directory '*'
 
 # Install CI tools using pip
-RUN pip install rapids-dependency-file-generator \
+RUN pip install rapids-dependency-file-generator==1 \
     && pip cache purge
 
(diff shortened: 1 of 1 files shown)


vyasr:
I believe we discussed this, but want to make sure: if a project wants to test out a newer version of dfg, they should be able to inject a `pip install rapids-dependency-file-generator==$VERSION` into the style check workflow and use the updated version without needing to update the image, correct?

ajschmidt8:
> I believe we discussed this, but want to make sure: if a project wants to test out a newer version of dfg, they should be able to inject a `pip install rapids-dependency-file-generator==$VERSION` into the style check workflow and use the updated version without needing to update the image, correct?

Yes, that will be implemented as an option to the relevant shared workflows in https://github.com/rapidsai/shared-action-workflows/

PR #31, opened 2023-01-03
Add libGL to image dependencies.

bdice:
Potential fix for OpenCV issues in https://github.com/rapidsai/cucim/pull/471.

(There are alternative fixes to consider, so this PR may be closed if another route is taken.)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -33,6 +33,7 @@ RUN \
           cuda-gdb-${PKG_CUDA_VER} \
           cuda-cudart-dev-${PKG_CUDA_VER} \
           cuda-cupti-dev-${PKG_CUDA_VER} \
+          libgl1 \
           wget \
         # ignore the build-essential package since it installs dependencies like gcc/g++
         # we don't need them since we use conda compilers, so this keeps our images smaller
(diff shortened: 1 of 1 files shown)


bdice:
We are adopting another solution here: https://github.com/rapidsai/cucim/pull/471/commits/6045729cb19ef08e65f573a21c41b974c2dba253

jakirkham:
Is there more context on why this is being closed and that option being chosen?

bdice:
@jakirkham I discussed on a call with @ajschmidt8. We decided to retain the same testing method that cucim used previously (`pip install opencv-python-headless`) because it was working in Jenkins (gpuCI).

To use conda packages of opencv (which require libGL), we'd have to install libGL into our CI images (which was not present in the old CI images either).

There are a couple discussions about creating "headless" opencv conda-forge packages that do not depend on libGL, which we would benefit from having:
- https://github.com/conda-forge/opencv-feedstock/issues/293
- https://github.com/conda-forge/opencv-feedstock/issues/321

"Headless" versions of opencv already exist on other channels such as `fastai`.

PR #32, opened 2023-01-09
Add `always_yes: true` to `.condarc` file

ajschmidt8:
Who is `conda` to make us second guess ourselves? Of course we want to install those dependencies.

This PR adds `always_yes: true` to our `.condarc` file so that `conda` won't prompt the user when performing `conda install` commands.


bdice (approved):
🐍 Sssssuper.

(For additional background -- the CI jobs run in noninteractive mode, and thus don't need this flag. The conda CLI just knows to accept "yes." However, this isn't true when running CI containers locally. When run locally, we don't want CI containers to prompt the user.)

PR #36, opened 2023-01-26
Update `sccache` bucket to `us-east-2`

ajschmidt8:
This PR updates the `sccache` environment variables to use the new `rapids-sccache-east` bucket, which is in the same region as the rest of our CI resources in the cloud.


ajschmidt8:
Keeping this PR as a draft until we actually create this bucket.

PR #47, opened 2023-03-21
WIP: native docker builds

sevagh:
WIP: building images on their native architecture.


axis.yaml
```diff
@@ -1,28 +1,30 @@
...
+  # skip ARM builds for CUDA 11.2.2
+  - CUDA_VER: 11.2.2
+    RUNNER_ARCH: "arm64"
 
 # Define the values used for the "latest" tag
 LATEST_VERSIONS:
-  CUDA_VER: "11.8.0"
-  PYTHON_VER: "3.10"
+  CUDA_VER: "12.0.1"
+  PYTHON_VER: "3.8"
```
ajschmidt8, line 29: the Python version here is probably fine to update.

We should leave the CUDA version as-is for now until all of RAPIDS is compatible with CUDA 12.

The `latest` tag is used by some jobs in CI (like [this](https://github.com/rapidsai/cudf/blob/bf18cea1461b63ffef32a5404b7552d1e90240e9/.github/workflows/pr.yaml#L86) `cudf` notebooks test job). So if we update this tag to a CUDA version that is not currently compatible with RAPIDS, it will likely break things.

ajschmidt8: Oh whoops. I see you're currently testing a reduced matrix :man_facepalming:.

Serves me right for commenting on a WIP PR. disregard the noise!

PR #53, opened 2023-05-23
Drop Python 3.8, add CUDA 12.1.1, update some versions.

bdice:
This PR drops images for Python 3.8 and adds images with CUDA 12.1.1. While CUDA 12.1 is not yet planned for our CI matrix, it would be useful to have these images for local testing.


ajschmidt8:
We'll need to add `12.1.1` to the https://github.com/rapidsai/mambaforge-cuda/ repo too.

Dockerfile
```diff
@@ -81,7 +81,7 @@ RUN rapids-mamba-retry install -y \
     gh \
     git \
     jq \
-    "sccache>=0.3.2" \
+    "sccache>=0.5.0" \
```
bdice, line 84: Running into issues with 0.5.0 -- temporarily pinning this. cc: @ajschmidt8
```suggestion
    "sccache==0.4.2" \
```

PR #56, opened 2023-07-13
Set latest image to use CUDA 12.0.1.

bdice:
This PR should be merged after https://github.com/rapidsai/shared-action-workflows/pull/101.


bdice:
Let's target this for early 23.10, so that we don't break anything unexpected right before the release.

bdice:
There was some discussion about whether we need tags like `cuda11-latest` and `cuda12-latest` but I would prefer to avoid that, and just have one `latest` tag.

ajschmidt8 (approved):

raydouglass (approved):

PR #61, opened 2023-08-21
Build multi-arch images natively

AyodeAwe:
PR updates the CI workflows to build multi-arch images natively. This would result in stable builds (as opposed to the occasionally unstable builds we have with QEMU)

The workflow does the following:
- Computes the required matrix values
- Builds the docker images for each variant of the computed matrix
- On branch builds and PRs, builds and pushes images and pushes multi-arch manifests. 
- Removes temporary docker images created during the multi-architectural build process.

PS: Changes is this PR were ported from this [one](https://github.com/rapidsai/cibuildwheel-imgs/pull/33).


.github/workflows/build-and-publish-images.yml
```diff
@@ -15,71 +15,80 @@ jobs:
   compute-matrix:
     runs-on: ubuntu-latest
     outputs:
-      LATEST_LINUX_VER: ${{ steps.latest-values.outputs.LATEST_LINUX_VER }}
-      LATEST_CUDA_VER: ${{ steps.latest-values.outputs.LATEST_CUDA_VER }}
-      LATEST_PYTHON_VER: ${{ steps.latest-values.outputs.LATEST_PYTHON_VER }}
       MATRIX: ${{ steps.compute-matrix.outputs.MATRIX }}
+      LATEST_LINUX_VER: ${{ steps.compute-matrix.outputs.LATEST_LINUX_VER }}
+      LATEST_CUDA_VER: ${{ steps.compute-matrix.outputs.LATEST_CUDA_VER }}
+      LATEST_PYTHON_VER: ${{ steps.compute-matrix.outputs.LATEST_PYTHON_VER }}
```
ajschmidt8, line 21: I think this change should be reverted. The `compute-matrix` step doesn't output these values.

AyodeAwe: Yeah that was a typo, should be `latest-values` not `compute-matrix`.

ajschmidt8 (approved):

PR #62, opened 2023-08-22
Fix `codecov` installation

ajschmidt8:
#61 added an explicit `TARGETPLATFORM` build argument, which broke our CI images.

`TARGETPLATFORM` is a build argument provided by `docker build` automatically: https://docs.docker.com/engine/reference/builder/#automatic-platform-args-in-the-global-scope.


ajschmidt8:
The `TARGETPLATFORM` argument is used to install `codecov` below:

- https://github.com/rapidsai/ci-imgs/blob/c1eedbae84890228e22c2a33519fe9a93b9fc81d/Dockerfile#L87-L106

PR #68, opened 2023-09-08
Add upstream info

AyodeAwe:
PR adds information about the triggering upstream job. Also adds the `workflow_dispatch` event. 

NOTE:
- `upstream_job` will be visible as an input parameter for the `publish` job. For instance, [here](https://github.com/rapidsai/ci-imgs/actions/runs/6124467698/job/16624547397#step:1:37).
- `upstream_repository` input is used to display the run name (in the actions page)


ajschmidt8 (approved):

ajschmidt8:
It's clear where `upstream_repository` is used, but can you add information here about where `upstream_job` will be visible for posterity?

PR #72, opened 2023-09-19
Publish `ci-conda` image tag

AyodeAwe:
Publish the existing `rapidsai/ci` images as both `rapidsai/ci` and `rapidsai/ci-conda`. This is part of the ongoing work to rename our `rapidsai/ci` images to `rapidsai/ci-conda`.

After updating references across our codebase from `rapidsai/ci` to `rapidsai/ci-conda`, a second PR will be opened reverting some changes in this PR, and making the existing setup publish the image as `rapidsai/ci-conda` and not as `rapidsai/ci` any longer.


matrix.yaml
```diff
@@ -65,4 +65,4 @@ exclude:
 
   # exclude ci for ubuntu18.04
```
ajschmidt8, line 66: can you update the comments in this file that reference `ci` to `ci-conda`?

it looks like there are a few

ajschmidt8 (commented):
flawless. looks good to me. i inspected the GHAs logs and it seems like things are working as expected.

Can you just update the README? then I think this PR is good to go.

ajschmidt8 (approved):

PR #86, opened 2023-10-26
Try adding Ubuntu 18.04 wheels

vyasr:
We need these to build newer manylinux wheels for cudf.


vyasr:
We're going to try and get proper glibc 2.28 images instead of these 2.24 ones that we'd have to retag wheels from.

PR #89, opened 2023-10-30
Add `which` to `ci-wheel`

jakirkham:
This was missing in the Rocky Linux 8 image. Though is often relied on by build scripts. So make sure `which` is always installed.

### DIFF
--- a/ci-wheel.Dockerfile
+++ b/ci-wheel.Dockerfile
@@ -30,7 +30,7 @@ RUN case "${LINUX_VER}" in \
         echo 'APT::Update::Error-Mode "any";' > /etc/apt/apt.conf.d/warnings-as-errors \
         && apt update -y \
         && apt install -y \
-          jq build-essential software-properties-common wget gcc zlib1g-dev libbz2-dev \
+          jq build-essential software-properties-common which wget gcc zlib1g-dev libbz2-dev \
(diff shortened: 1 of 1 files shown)


bdice (approved):

jakirkham:
> Please consider updating the PR description to explain why this package is needed, for future reference.

Done

ci-wheel.Dockerfile
```diff
@@ -30,7 +30,8 @@ RUN case "${LINUX_VER}" in \
         echo 'APT::Update::Error-Mode "any";' > /etc/apt/apt.conf.d/warnings-as-errors \
         && apt update -y \
         && apt install -y \
-          jq build-essential software-properties-common which wget gcc zlib1g-dev libbz2-dev \
+          debianutils build-essential software-properties-common \
```
jakirkham, line 33: Need to use [`debianutils` to get `which` on Ubuntu]( https://packages.ubuntu.com/focal/amd64/debianutils/filelist )

jakirkham:
Yeah just trying to make sure it is always there. We already were relying on it. So it seems better to document that requirement and added it consistently across images.

Otherwise subtle differences between the images could be a source of issues down the road.

vyasr (approved):

PR #96, opened 2023-11-28
Drop CUDA 11.2 and 11.5, add Python 3.11

ajschmidt8:
This is a follow-up to https://github.com/rapidsai/miniforge-cuda/pull/55.

### DIFF
--- a/matrix.yaml
+++ b/matrix.yaml
@@ -7,7 +7,6 @@ PYTHON_VER:
   - "3.9"
   - "3.10"
   - "3.11"
-  - "3.12"
 LINUX_VER:
   - "ubuntu18.04"
   - "ubuntu20.04"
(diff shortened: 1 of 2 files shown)


bdice:
Conda is failing to solve `boa` with `python 3.12` because it can't find a `conda-build` package for Python 3.12. The current state of the conda-forge Python 3.12 migration will need https://github.com/conda-forge/lief-feedstock/pull/41 in order to unblock conda-build packages for Python 3.12, which should solve this.

For now, I am going to remove Python 3.12 from this PR and we can revisit that at a later time.

matrix.yaml
```diff
@@ -1,13 +1,13 @@
...
   - "11.4.3"
-  - "11.5.2"
   - "11.8.0"
   - "12.0.1"
   - "12.1.1"
 PYTHON_VER:
   - "3.9"
   - "3.10"
+  - "3.11"
+  - "3.12"
```
bdice, line 10:
```suggestion
```

PR #100, opened 2023-12-18
remove unnecessary build arg in ci-wheel Dockerfile

jameslamb:

Removes one unnecessary build argument in the `ci-wheel` Dockerfile.

Since build arg `REAL_ARCH` isn't used in this `FROM`, specifying it before that has no effect:

https://github.com/rapidsai/ci-imgs/blob/a02c2e8b73ecd3f1dd0d1e18ee944229a98729b4/ci-wheel.Dockerfile#L3-L5

Per https://docs.docker.com/engine/reference/builder/#scope

> *An ARG instruction goes out of scope at the end of the build stage where it was defined.*


ci-wheel.Dockerfile
```diff
@@ -1,14 +1,13 @@
...
-ARG REAL_ARCH=x86_64
 
 ARG BASE_IMAGE=nvcr.io/nvidia/cuda:${CUDA_VER}-devel-${LINUX_VER}
 FROM ${BASE_IMAGE}
 
 ARG CUDA_VER
 ARG LINUX_VER
 ARG CPU_ARCH
-ARG REAL_ARCH
+ARG REAL_ARCH=x86_64
```
vyasr, line 10: I think we always pass this variable in, so maybe better not to have a default?

vyasr (commented):
LGTM. Seems like the `CPU_ARCH` is used once (for downloading sccache)

vyasr:
Also related question, is it possible to infer `CPU_ARCH` from `REAL_ARCH` or vice versa? I always forget the fine points of the differences between things like x86_64 and amd64.

ajschmidt8 (approved):

ajschmidt8:
> Also related question, is it possible to infer `CPU_ARCH` from `REAL_ARCH` or vice versa? I always forget the fine points of the differences between things like x86_64 and amd64.

There's likely some additional refactoring that can be done here, but we can save that for the future. Divye and I had a chat about this a while ago too.

For now, we'll merge this change as it is.

Thanks, James!

PR #101, opened 2023-12-19
add CUDA 12.2

jameslamb:

Blocked by https://github.com/rapidsai/miniforge-cuda/pull/56.

* adds CUDA 12.2 images for `ci-conda`, `ci-wheel`, and `citestwheel`
* drops builds of CUDA 12.1 images

### How is it safe to stop generating CUDA 12.1 images?

According to the description in #53, those 12.1.1 images were added for "local testing". It seems they're not being used for anything other than that.

This project is currently only generating `rapidsai/ci-conda` images for CUDA 12.1.

https://github.com/rapidsai/ci-imgs/blob/a02c2e8b73ecd3f1dd0d1e18ee944229a98729b4/matrix.yaml#L31-L39

Every use of `rapidsai/ci-conda` I could find ([GitHub search](https://github.com/search?q=org%3Arapidsai+ci-conda&type=code&p=1)) is using either the CUDA 11.8.0 images directly, or `:latest`.

`:latest` for `ci-conda` is `rapidsai/ci-conda:cuda12.0.1-ubuntu22.04-py3.10`, and this PR doesn't change that:

https://github.com/rapidsai/ci-imgs/blob/a02c2e8b73ecd3f1dd0d1e18ee944229a98729b4/latest.yaml#L1-L5

### Notes for Reviewers

12.2.2 is the latest tag I see for the 12.2.x series at https://hub.docker.com/r/nvidia/cuda/tags?page=1&name=12.2.


matrix.yaml
```diff
@@ -28,15 +28,11 @@ exclude:
...
-  # exclude citestwheel and ci-wheel for cuda versions other than 11.8.0 and 12.0.1
+  # exclude citestwheel and ci-wheel for cuda versions other than 11.8.0, 12.0.1, and 12.2.2
   - CUDA_VER: "11.4.3"
     IMAGE_REPO: "citestwheel"
   - CUDA_VER: "11.4.3"
     IMAGE_REPO: "ci-wheel"
-  - CUDA_VER: "12.1.1"
-    IMAGE_REPO: "citestwheel"
-  - CUDA_VER: "12.1.1"
-    IMAGE_REPO: "ci-wheel"
```
ajschmidt8, line 39: I think these exclusions should be updated to include `12.2.2`.

@vyasr can comment, but I think for wheels we want to build/test on the earliest CUDA images.

jameslamb: I chose not to add these exclusions based on this statement in https://github.com/rapidsai/build-planning/issues/7:

> *Modify the [wheel shared-workflows](https://github.com/rapidsai/shared-workflows/tree/branch-24.02/.github/workflows) to use the new images in building wheels. These jobs [should be set up to continue on error](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idcontinue-on-error)*

Won't `ci-wheel` and `citestwheel` images with CUDA 12.2 be necessary to do that?

ajschmidt8: I'll let Vyas comment. He knows the wheel stuff better than me.

At the very least, the quote you mentioned applies to the conda images. Not sure about the wheels.

jameslamb: > I'll let Vyas comment.

hmmm ok, yeah I'll wait for that.

I guess those exclusions were also covering the fact that there aren't Ubuntu 18.04 `nvidia/cuda` images CUDA 12.1 ([DockerHub search](https://hub.docker.com/r/nvidia/cuda/tags?page=1&name=12.1.1-devel-ubuntu)) or 12.2 ([DockerHub search](https://hub.docker.com/r/nvidia/cuda/tags?page=1&name=12.2.2-devel-ubuntu)) on DockerHub, but there are for 12.0 ([DockerHub search](https://hub.docker.com/r/nvidia/cuda/tags?page=1&name=12.0.1-devel-ubuntu)).

https://github.com/rapidsai/ci-imgs/actions/runs/7277784010?pr=101

image

bdice: Yes, we'll need to exclude Ubuntu 18.04 for CUDA 12.2. Discussed here: https://github.com/rapidsai/shared-workflows/pull/166#discussion_r1439758006

bdice: I attempted that exclusion in d51e8d926f470801da00e81973f0a8eea527fd59. Feel free to edit!

bdice: To the original question from @ajschmidt8:

> I think these exclusions should be updated to include `12.2.2`.

I think we want:
- build on "newest CUDA minor version from each supported major version" = (11.8, 12.2)
- test on "newest CUDA minor version from previous major version" = 11.8; "oldest CUDA minor version from current major version" = 12.0, and "newest CUDA minor version (that RAPIDS supports) from current major version" = 12.2.

The testing part aligns with my comment here about previous policy discussions around the time we introduced CUDA 12.0 support: https://github.com/rapidsai/shared-workflows/pull/166#discussion_r1439743881

(We need to figure out how to document our _reasons_ and not just our results...)

jameslamb: > I attempted that exclusion in https://github.com/rapidsai/ci-imgs/commit/d51e8d926f470801da00e81973f0a8eea527fd59

Works for me, thanks for that!

> I think we want...

@bdice Given this, do you think I could safely remove cuda11.4 builds from this repo?

bdice: > do you think I could safely remove cuda11.4 builds from this repo?

Yes, probably, but I'd like to check with the ops team and separate that into a new PR.

ajschmidt8 (approved):

matrix.yaml
jameslamb: Ok great, I'll do that.

jameslamb: eh well, maybe that's not as simple as I thought. I'll defer dropping CUDA 11.4 for now.

I see it's still tested against.

https://github.com/rapidsai/shared-workflows/blob/91799a905608f4c57d7fe65b92ce9a261249e635/.github/workflows/conda-cpp-tests.yaml#L66

https://github.com/rapidsai/shared-workflows/blob/91799a905608f4c57d7fe65b92ce9a261249e635/.github/workflows/conda-cpp-tests.yaml#L69

https://github.com/rapidsai/shared-workflows/blob/91799a905608f4c57d7fe65b92ce9a261249e635/.github/workflows/conda-python-tests.yaml#L74-L76

https://github.com/rapidsai/shared-workflows/blob/91799a905608f4c57d7fe65b92ce9a261249e635/.github/workflows/conda-python-tests.yaml#L74-L76

bdice (approved):

jameslamb:
Thank you both! I'll merge this and move on to testing with `shared-workflows` today.

matrix.yaml
vyasr: Although this PR has already been merged, I'm following up for the record since I was tagged a handful of times. We ended up with mostly the right thing here. The current matrix is fine, but we could add some extra exclusions to avoid building unnecessary images. In particular, we could exclude CUDA 12.0/12.1 ci-wheel images since we will always be building on the latest (12.2), and we can probably exclude CUDA 12.1 citestwheel images since we only need to test earliest and latest.

bdice: I'd wait to drop 12.0 ci-wheel images until we complete https://github.com/rapidsai/build-planning/issues/7.

I agree we don't need CUDA 12.1 images for anything, those should be safe to drop.

jameslamb: >  I'd wait to drop 12.0 ci-wheel images until we complete https://github.com/rapidsai/build-planning/issues/7.

This is what I was thinking too. Most projects still haven't been cut over to building on 12.2 and therefore need the 12.0 images for building.

vyasr: Yup all good, I was making these comments from the perspective of this PR having already been merged and what future changes we would want to see.

PR #102, opened 2023-12-20
make build-arg passing stricter

jameslamb:

* removes some defaults in `ARG` statements in the project's Dockerfiles
* replaces other defaults with placeholder values that make loud errors more likely

### Benefits of this change

* reduces the risk of accidentally falling back to older versions of those things (in the case where build args are not passed through correctly)
* makes the flow of configuration clearer
    - e.g. makes it clearer that `matrix.yaml` is the only place that holds information about what operating systems, Python versions, and CUDA versions images are built for

### Example

Consider the following:

```shell
docker-buildx build \
    -f ./ci-conda.Dockerfile \
    -t rapidsai/ci-conda:local \
    ./context
```

On `main`, that'll happily build a base image from `rapidsai/miniforge-cuda:cuda11.8.0-base-ubuntu22.04-py3.10`.

As of this PR, that'd result in a loud error indicating that the expected arguments hadn't been passed through.

> *ERROR: failed to solve: rapidsai/miniforge-cuda:cuda-notset-base-notset-py-notset: docker.io/rapidsai/miniforge-cuda:cuda-notset-base-notset-py-notset: not found*


ci-conda.Dockerfile
```diff
@@ -1,6 +1,6 @@
-ARG CUDA_VER=11.8.0
-ARG LINUX_VER=ubuntu22.04
-ARG PYTHON_VER=3.10
+ARG CUDA_VER=-notset
+ARG LINUX_VER=notset
+ARG PYTHON_VER=-notset
```
bdice, line 3: I'd drop the extra dashes. It's a bit more confusing than needed for the relatively minor benefit of readable words in errors.
```suggestion
ARG CUDA_VER=notset
ARG LINUX_VER=notset
ARG PYTHON_VER=notset
```

bdice (approved):

ci-conda.Dockerfile
jameslamb: Ok sure, I'm indifferent 🤷🏻 

You picked up on exactly why I had them... so you get `cuda-notset-*` instead of `cudanotset`.

ajschmidt8 (approved):

PR #108, opened 2024-02-15
Configure regex manager for unreachable dependencies

AyodeAwe:
This PRs enables `renovate` to locate dependencies that aren't findable using conventional `renovate` syntax. See the [regex manager docs](https://docs.renovatebot.com/modules/manager/regex/) for details. 

The `fix-renovate-args.sh` script is also written to accomodate building the dockerfiles locally.


AyodeAwe:
Changes were tested - the following PRs were created:

https://github.com/AyodeAwe/literate-octo-potato/pull/31
https://github.com/AyodeAwe/literate-octo-potato/pull/30
https://github.com/AyodeAwe/literate-octo-potato/pull/29
https://github.com/AyodeAwe/literate-octo-potato/pull/28
https://github.com/AyodeAwe/literate-octo-potato/pull/27

ci-conda.Dockerfile
```diff
@@ -1,12 +1,17 @@
 ARG CUDA_VER=notset
 ARG LINUX_VER=notset
 ARG PYTHON_VER=notset
+ARG YQ
+
+FROM mikefarah/yq:${YQ} as yq
```
ajschmidt8, line 6: why was this `FROM` statement added?

for consistency with the `amazon/aws-cli` `COPY` statements, we should revert this back to how it was before.

renovate.yaml
ajschmidt8, line 1: the versions in this file shouldn't contain a `v` prefix. the `v` character isn't part of the version number.

ci/fix-renovate-args.sh
```diff
@@ -0,0 +1,44 @@
+#!/bin/bash
+
```
ajschmidt8, line 2:
```suggestion
#!/bin/bash
set -euo pipefail
```

ci/fix-renovate-args.sh
```diff
@@ -0,0 +1,44 @@
...
+else
+  eval "docker build -f \"$DOCKERFILE_PATH\" . \
+    --build-arg CUDA_VER=\"$CUDA_VER\" \
+    --build-arg LINUX_VER=\"$LINUX_VER\" \
+    --build-arg PYTHON_VER=\"$PYTHON_VER\" \
+    --build-arg CPU_ARCH=\"$ARCH\" \
+    --build-arg REAL_ARCH=\"$(arch)\" \
+    --build-arg MANYLINUX_VER=\"$MANYLINUX_VER\" \
+    $BUILD_ARGS"
+fi
```
ajschmidt8, line 44: I think this file might need some more thought...

It looks like in GHAs, it will echo build arguments but outside of CI it will perform a `docker build`.

I'm not sure we want a single file that performs two totally different functions depending on whether it's run in GHAs or not.

ajschmidt8: Yes. It's the implementation that I think needs to be reconsidered.

As a contributor to the repository, it's not intuitive to me that running a script called `fix-renovate-args.sh` will perform a Docker build.

ci-conda.Dockerfile
AyodeAwe: From my testing, the `YQ` argument cannot be interpolated into the following statement:

https://github.com/rapidsai/ci-imgs/blob/main/ci-conda.Dockerfile#L158

as in:

`COPY --from=mikefarah/yq:${YQ} /usr/bin/yq /usr/local/bin/yq`

ajschmidt8: ok, that's fine. we can leave it.

for consistency, let's add a version to the `aws` CLI image and add a `FROM` statement for that as well.

ci/fix-renovate-args.sh
ajschmidt8: I think a better implementation could be a Makefile with a `build` target.

The `build` target will run `docker build` and call a script under the hood to parse `versions.yaml` and inject the versions as build arguments. The target can also pass any additional positional parameters to the `docker build`  command.

An example might look like...

```sh
make build -f Dockerfile context/
```

ajschmidt8: alternatively, a copy-pasteable README snippet is probably also sufficient. This would avoid the need to create a wrapper script.

e.g:

```sh
docker build $(ci/compute-build-args.sh) -f Dockerfile context/
```

renovate.yaml
```diff
@@ -0,0 +1,10 @@
+# renovate: datasource=github-releases depName=mozilla/sccache
+SCCACHE: v0.7.6
```
raydouglass, line 2: I think these names should match what we use elsewhere (eg `CUDA_VER`, `PYTHON_VER`, etc).

I think looking at the Dockfiles with `ARG YQ` and `ARG CODECOV` doesn't necessarily capture what they are for. But `YQ_VER` and `CODECOV_VER` do.

README.md
```diff
@@ -9,3 +9,11 @@ This repository includes the following CI images for RAPIDS:
 ## `latest` tag
 
 The `latest` image tags are controlled by the values in `latest.yaml`.
+
+### Building the dockerfiles locally
+To easily build the dockerfiles locally, you may use the following snippets:
+```
+docker build $(ci/compute-build-args.sh) -f ci-conda.Dockerfile context/
```
ajschmidt8, line 16:
```suggestion
```sh
export LINUX_VER=ubuntu22.04
export CUDA_VER=12.2
export PYTHON_VER=3.11
export ARCH=amd64
docker build $(ci/compute-build-args.sh) -f ci-conda.Dockerfile context/
```

If the default values are removed, we should probably update the snippet to instruct people to define the values that they want to build for.

ci-conda.Dockerfile
```diff
@@ -1,12 +1,21 @@
...
 ARG LINUX_VER=notset
 ARG PYTHON_VER=notset
+ARG YQ_VER
+ARG AWS_CLI_VER
+ARG SCCACHE_VER
+
+FROM mikefarah/yq:${YQ_VER} as yq
+
+FROM amazon/aws-cli:${AWS_CLI_VER} as aws-cli
+
```
ajschmidt8, line 11: only `ARG`s that are used in the `FROM` directives need to be specified at the top.

the remaining `ARG`s should come after the `FROM` statements, otherwise they won't be recognized throughout the rest of the Dockerfile.

These docs outline this behavior: https://docs.docker.com/reference/dockerfile/#understand-how-arg-and-from-interact.

ci-conda.Dockerfile
```diff
@@ -1,12 +1,21 @@
...
 FROM rapidsai/miniforge-cuda:cuda${CUDA_VER}-base-${LINUX_VER}-py${PYTHON_VER}
 
 ARG TARGETPLATFORM
 ARG CUDA_VER
 ARG LINUX_VER
 ARG PYTHON_VER
+ARG CODECOV_VER
+ARG SCCACHE_VER
```
ajschmidt8, line 18: In the `ci-wheel.Dockerfile` file, the `ARG` statements are placed right above the `RUN` statement where they are used.

This file is inconsistent with that file.

Let's move these `ARG`s above the `RUN` statements for consistency with `ci-wheel.Dockerfile`.

README.md
```diff
@@ -9,3 +9,15 @@ This repository includes the following CI images for RAPIDS:
 ## `latest` tag
 
 The `latest` image tags are controlled by the values in `latest.yaml`.
+
+## Building the dockerfiles locally
+To easily build the dockerfiles locally, you may use the following snippets:
+```sh
```
ajschmidt8, line 15:
```suggestion
## Building the dockerfiles locally

To easily build the dockerfiles locally, you may use the following snippets:

```sh
```

Markdown formatters generally put newlines between these elements. Let's add the newlines since we'll be doing one last round of edits.

README.md
```diff
@@ -9,3 +9,15 @@ This repository includes the following CI images for RAPIDS:
...
+export LINUX_VER=ubuntu22.04
+export CUDA_VER=12.2
```
ajschmidt8, line 17:
```suggestion
export CUDA_VER=12.2.2
```

`12.2` isn't a valid CUDA version for our images.

ajschmidt8 (approved):

PR #111, opened 2024-02-21
Use CUDA 12.2 for latest tag.

bdice:
This PR updates the "latest" tag to use CUDA 12.2 images.


jameslamb (approved):
blegh sorry I missed this one in the list at https://github.com/rapidsai/build-planning/issues/7#issuecomment-1957227696.

Thank you for adding it there and for doing this.

PR #112, opened 2024-02-22
Add conda-merge to ci-conda images.

bdice:
I'm experimenting with a change to conda CI workflows that would install test dependencies and library packages at the same time, rather than creating an environment with `rapids-dependency-file-generator` containing only test dependencies and then later adding the library packages. See: https://github.com/rapidsai/cuml/pull/5781

To do this in the simplest way, I would like to add `conda-merge` to the `ci-conda` images. We can merge this change to CI images if the approach in the cuML PR seems like an improvement.

### DIFF
--- a/ci-conda.Dockerfile
+++ b/ci-conda.Dockerfile
@@ -111,6 +111,7 @@ RUN <<EOF
 rapids-mamba-retry install -y \
   anaconda-client \
   boa \
+  conda-merge \
   gettext \
   gh \
   git \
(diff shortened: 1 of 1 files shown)


bdice:
[Maybe there's a better solution](https://github.com/rapidsai/cuml/pull/5781#issuecomment-1958498274) that uses `rapids-dependency-file-generator`. Let's hold off on merging for now.

bdice:
Closing in favor of #113.

PR #113, opened 2024-02-22
Update to rapids-dependency-file-generator v1.9.0

bdice:
This PR builds CI images with the latest rapids-dependency-file-generator. The change in this PR is small - it just makes the pinnings match between `ci-conda` and `ci-wheel`, but it'll fetch the latest version when rebuilding.


jameslamb (approved):

raydouglass (approved):

jakirkham:
Thanks all! 🙏

PR #117, opened 2024-02-28
fix wrong fileMatch target

AyodeAwe:
`fileMatch` key is currently incorrectly configured. This PR fixes it.


ajschmidt8 (approved):

ajschmidt8:
canceled CI and will admin merge since this change isn't tested in GHAs.

PR #125, opened 2024-03-01
Use Python 3.11 in latest tag.

bdice:
Updates `latest` tag to use Python 3.11.

xref: https://github.com/rapidsai/build-planning/issues/3


raydouglass (approved):

jakirkham:
Thank you both! 🙏

PR #130, opened 2024-03-27
Install rapids-dependency-file-generator from Conda

KyleFromNVIDIA:
Now that Conda packages are available for
`rapids-dependency-file-generator`, use those instead of the Pip packages.

### DIFF
--- a/ci-conda.Dockerfile
+++ b/ci-conda.Dockerfile
@@ -106,6 +106,7 @@ RUN wget https://github.com/rapidsai/gha-tools/releases/latest/download/tools.ta
 # Install CI tools using mamba
 RUN <<EOF
 rapids-mamba-retry install -y \
+  -c rapidsai
   anaconda-client \
   boa \
   dunamai \
--- a/ci-conda.Dockerfile
+++ b/ci-conda.Dockerfile
@@ -103,10 +103,16 @@ EOF
 RUN wget https://github.com/rapidsai/gha-tools/releases/latest/download/tools.tar.gz -O - \
   | tar -xz -C /usr/local/bin
 
+# Create condarc file from env vars
+ENV RAPIDS_CONDA_BLD_ROOT_DIR=/tmp/conda-bld-workspace
+ENV RAPIDS_CONDA_BLD_OUTPUT_DIR=/tmp/conda-bld-output
+COPY condarc.tmpl /tmp/condarc.tmpl
+RUN cat /tmp/condarc.tmpl | envsubst | tee /opt/conda/.condarc; \
+    rm -f /tmp/condarc.tmpl
+
 # Install CI tools using mamba
 RUN <<EOF
 rapids-mamba-retry install -y \
-  -c rapidsai \
   anaconda-client \
   boa \
   dunamai \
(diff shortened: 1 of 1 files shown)


ci-conda.Dockerfile
```diff
@@ -164,10 +165,6 @@ RUN cat /tmp/condarc.tmpl | envsubst | tee /opt/conda/.condarc; \
 
 RUN /opt/conda/bin/git config --system --add safe.directory '*'
 
-# Install CI tools using pip
-RUN pip install dunamai "rapids-dependency-file-generator==1.*" \
```
ajschmidt8, line 168: was `dunamai` intentionally removed here? I think we still need that

ajschmidt8: though it looks like that can be fetched from conda too:

- https://anaconda.org/conda-forge/dunamai

ci-conda.Dockerfile
```diff
@@ -111,7 +111,8 @@ rapids-mamba-retry install -y \
   gettext \
   git \
   jq \
-  "python=${PYTHON_VERSION}.*=*_cpython"
+  "python=${PYTHON_VERSION}.*=*_cpython" \
+  rapids-dependency-file-generator
```
ajschmidt8, line 115: we should still keep this pinned to version `1.*`

ci-conda.Dockerfile
KyleFromNVIDIA: Updated to get dunamai from Conda.

ajschmidt8 (approved):
:fire: :fire: :fire:

vyasr (approved):

ajschmidt8 (changes requested):
the `rapidsai` channel is added in these Docker steps:

- https://github.com/rapidsai/ci-imgs/blob/26fd8513c8a0582a478f611b8941ab80a629db8e/ci-conda.Dockerfile#L161-L166

Can we move those steps above the `conda install` commands to avoid having to specify the `-c rapidsai` flag?

KyleFromNVIDIA:
Done

ajschmidt8:
> the `rapidsai` channel is added in these Docker steps:
> 
> * https://github.com/rapidsai/ci-imgs/blob/26fd8513c8a0582a478f611b8941ab80a629db8e/ci-conda.Dockerfile#L161-L166
> 
> Can we move those steps above the `conda install` commands to avoid having to specify the `-c rapidsai` flag?

Ah. Hmm. I forgot that the `envsubst` command in those lines comes from this `gettext` package that's installed via `conda`:

- https://github.com/rapidsai/ci-imgs/blob/26fd8513c8a0582a478f611b8941ab80a629db8e/ci-conda.Dockerfile#L113

If I recall correctly, I installed it from `conda` instead of a system package manager because the size of the package and it's required dependencies were significantly smaller.

I wonder if there's a better way to do this.

If not, I suppose we can keep the `-c` flag.

KyleFromNVIDIA:
I've installed `gettext` in a separate step.

ajschmidt8 (approved):

raydouglass (approved):

PR #134, opened 2024-04-23
Use Conda to install additional packages

jakirkham:
Fix install issues resulting in CI failures where PyPy is getting pulled in unexpectedly


msarahan:
I think the purpose here was to try to work around the confusion with some of the images having Pypy instead of cpython, right? Was there another reason here?

jakirkham:
Yes that was the main motivator

There may be secondary considerations between these 2 options, but fixing CI is the goal

msarahan:
In that case, I think https://github.com/rapidsai/miniforge-cuda/pull/52 is a better fix that serves the same purpose. OK with closing this?

jakirkham:
Sure we can always revisit if needed

PR #139, opened 2024-05-14
Remove centos and ubuntu18.04 from ci-imgs matrix

AyodeAwe:
`centos7` and `ubuntu18.04` are no longer supported across RAPIDS starting from `24.06`. PR removes any combination referencing these from the images build matrix.

Also, as RAPIDS will no longer build `manylinux_2_17` tags for its wheels, this PR updates the `build-image` workflow accordingly.

### DIFF
--- a/ci/compute-build-args.sh
+++ b/ci/compute-build-args.sh
@@ -3,7 +3,6 @@ set -euo pipefail
 
 MANYLINUX_VER="manylinux_2_17"
 if [[
-  "${LINUX_VER}" == "ubuntu18.04" ||
   "${LINUX_VER}" == "ubuntu20.04"
 ]]; then
   MANYLINUX_VER="manylinux_2_31"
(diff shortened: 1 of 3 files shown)


ci/compute-build-args.sh
```diff
@@ -3,7 +3,6 @@ set -euo pipefail
 
 MANYLINUX_VER="manylinux_2_17"
```
AyodeAwe, line 4: @bdice do we want to update this to `manylinux_2_28`?

bdice: Yes, but an even better solution would be to error if we’re using an image not listed below (Ubuntu 20.04 or Rocky 8). In my understanding, we must know and specify the correct glibc version for the OS here.

bdice (approved):

bdice:
Follow-up for miniforge-cuda: https://github.com/rapidsai/miniforge-cuda/pull/66

PR #147, opened 2024-06-04
install new tool codecov cli from source

AyodeAwe:
Implements a workaround to handle the issue mentioned here: https://github.com/rapidsai/ci-imgs/pull/142#issue-2328451536

How I tested:
- The problematic GLIBC issue was initially discovered in a pynvjitlink PR and described [here](https://github.com/rapidsai/ci-imgs/pull/142#issue-2328451536).
- So, I have tested that we don't get the issue with the source dist installation alternative, completed in another `pynvjitlink` PR [here](https://github.com/rapidsai/pynvjitlink/pull/87). I've done this by running the `pynvjitlink` conda tests in a modified/extended `rapidsai/ci-conda` image where I have installed the tool by source.
- And a second verification [here](https://github.com/rapidsai/literate-octo-potato/pull/827) (see the green checks in the PR) that the commands set in shared-workflows ([PR](https://github.com/rapidsai/shared-workflows/pull/215/files#diff-37f487709ba4c453085d35f9281dfc3cd3e6de4c3eb0f363b9a144fd38caccddR150-R154)) work as desired. This PR also runs against the same modified image (which has the tool installed by source)


ci-conda.Dockerfile
```diff
@@ -149,26 +153,16 @@ mv gh_*/bin/gh /usr/local/bin
...
-curl -Os --max-time 10 --retry 5 ${codecov_url}.SHA256SUM
-curl -Os --max-time 10 --retry 5 ${codecov_url}.SHA256SUM.sig
-
-gpgv codecov.SHA256SUM.sig codecov.SHA256SUM
-shasum -a 256 -c codecov.SHA256SUM
-chmod +x codecov
-mv codecov /usr/local/bin
-rm -f codecov.SHA256SUM codecov.SHA256SUM.sig
+# temporary workaround for discovered codecov binary install issue. See rapidsai/ci-imgs/issues/142
+wget "https://files.pythonhosted.org/packages/96/73/e18aaee2b3638528cfbece0615c34a59489f9063413744a31149558a0645/codecov-cli-${CODECOV_VER}.tar.gz"
```
bdice, line 160: Can we use `pip install` instead? We should not need to download the source distribution and install it manually.

bdice (changes requested):

ci-conda.Dockerfile
```diff
@@ -149,26 +153,11 @@ mv gh_*/bin/gh /usr/local/bin
...
-curl -Os --max-time 10 --retry 5 ${codecov_url}.SHA256SUM
-curl -Os --max-time 10 --retry 5 ${codecov_url}.SHA256SUM.sig
-
-gpgv codecov.SHA256SUM.sig codecov.SHA256SUM
-shasum -a 256 -c codecov.SHA256SUM
-chmod +x codecov
-mv codecov /usr/local/bin
-rm -f codecov.SHA256SUM codecov.SHA256SUM.sig
+# temporary workaround for discovered codecov binary install issue. See rapidsai/ci-imgs/issues/142
+pip install --no-binary codecov-cli codecov-cli==${CODECOV_VER}
```
bdice, line 160: Do we have to supply `--no-binary`? I think we would prefer to get the binary (wheel) on amd64 and let `pip` fall back to using the source distribution on arm64.
```suggestion
pip install codecov-cli==${CODECOV_VER}
```

versions.yaml
```diff
@@ -3,7 +3,7 @@ SCCACHE_VER: 0.7.7
 # renovate: datasource=github-releases depName=cli/cli
 GH_CLI_VER: 2.51.0
 # renovate: datasource=github-releases depName=codecov/uploader
```
bdice, line 5: We need to point renovate to the new source, which comes from PyPI. I don't know the proper syntax for this, but here is my best guess. Please verify this!
```suggestion
# renovate: datasource=pypi depName=codecov-cli
```

versions.yaml
```diff
@@ -3,7 +3,7 @@ SCCACHE_VER: 0.7.7
...
-CODECOV_VER: 0.7.2
+CODECOV_VER: 0.6.0
```
bdice, line 6: There is a new release: https://pypi.org/project/codecov-cli/#history
```suggestion
CODECOV_VER: 0.7.2
```

bdice (approved):
Please fix the renovate source and version, then this should be good to go.

versions.yaml
AyodeAwe: Latest version [here](https://pypi.org/project/codecov-cli/) reads `0.7.1`

versions.yaml
AyodeAwe: Yes this seems correct ([src](https://docs.renovatebot.com/modules/datasource/pypi/))

bdice (approved):
Thanks for catching my typo on the version and verifying the Renovate config. Looks good to me.

PR #154, opened 2024-06-28
remove ucx from wheel images

jameslamb:
As a result of https://github.com/rapidsai/build-planning/issues/57, it should no longer be necessary to install openucx in the wheel images here.

Projects depending on UCX in their CI all now get it from either the `ucx` conda-forge package or the `libucx` wheels published from https://github.com/rapidsai/ucx-wheels.

This proposes removing UCX builds in the wheel images. Benefits of that:

* reduces build times and complexity
* allows for some simplifications to CI in `ucx-py` and `ucxx` (e.g. https://github.com/rapidsai/ucxx/pull/226#discussion_r1583657525)

### DIFF
--- a/ci-wheel.Dockerfile
+++ b/ci-wheel.Dockerfile
@@ -133,31 +133,6 @@ EOF
 # Set AUDITWHEEL_* env vars for use with auditwheel
 ENV AUDITWHEEL_POLICY=${POLICY} AUDITWHEEL_ARCH=${REAL_ARCH} AUDITWHEEL_PLAT=${POLICY}_${REAL_ARCH}
 
-
-# Install ucx
-ARG UCX_VER=notset
-RUN <<EOF
-mkdir -p /ucx-src
-cd /ucx-src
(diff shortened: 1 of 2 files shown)


pentschev (approved):
There are several build issues that look unrelated, perhaps rate-limit issues:

```
 2 warnings found (use --debug to expand):
 - FromAsCasing: 'as' and 'FROM' keywords' casing do not match (line 7)
 - FromAsCasing: 'as' and 'FROM' keywords' casing do not match (line 9)
ERROR: toomanyrequests: Too Many Requests (HAP429).
```

Other than that LGTM, thanks @jameslamb !

jameslamb:
I just triggered a re-run of the failing jobs, I strongly suspect that'll fix it.

pentschev:
> I just triggered a re-run of the failing jobs, I strongly suspect that'll fix it.

Indeed it fixed. Thanks James!

vyasr (approved):

PR #163, opened 2024-07-16
Install rapids-dependency-file-generator in test-wheel image

seberg:
To allow testing the oldest dependencies, we wish to encode these into the `dependencies.yaml` file.  To generate them at test time we then need `rapids-dependency-file-generator` to be installed and it seems easier to do this here.

xref https://github.com/rapidsai/build-planning/issues/81

*NOTE: I haven't tested this locally yet, the README snippet didn't run right away for me adding the equivalent [here](https://github.com/rapidsai/rmm/pull/1613/files#diff-65e357e2da7e1573d93ff3794f8846d2517c42004ebc5aa5dafc1e4d02105559R15) worked and the build log looks right*

### DIFF
--- a/citestwheel.Dockerfile
+++ b/citestwheel.Dockerfile
@@ -56,6 +56,10 @@ ENV PATH="/pyenv/versions/${PYTHON_VER}/bin/:$PATH"
 COPY --from=aws-cli /usr/local/aws-cli/ /usr/local/aws-cli/
 COPY --from=aws-cli /usr/local/bin/ /usr/local/bin/
 
+# Install rapids-dependency-file-generator
+RUN pyenv global ${PYTHON_VER} && python -m pip install rapids-dependency-file-generator \
+  && pyenv rehash
+
 # Install latest gha-tools
 RUN wget https://github.com/rapidsai/gha-tools/releases/latest/download/tools.tar.gz -O - \
   | tar -xz -C /usr/local/bin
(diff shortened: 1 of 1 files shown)


citestwheel.Dockerfile
```diff
@@ -56,6 +56,10 @@ ENV PATH="/pyenv/versions/${PYTHON_VER}/bin/:$PATH"
 COPY --from=aws-cli /usr/local/aws-cli/ /usr/local/aws-cli/
 COPY --from=aws-cli /usr/local/bin/ /usr/local/bin/
 
+# Install rapids-dependency-file-generator
+RUN pyenv global ${PYTHON_VER} && python -m pip install rapids-dependency-file-generator \
```
jameslamb, line 60:
```suggestion
RUN pyenv global ${PYTHON_VER} && python -m pip install 'rapids-dependency-file-generator>=1.14.0,<2.0.0a0' \
```

Let's put an upper bound on this, to avoid breaking CI if a 2.0 is ever released. As is done for the other CI images, like: 

https://github.com/rapidsai/ci-imgs/blob/c57ffbdccfbd4ace85986237b4b6d441624bfcde/ci-conda.Dockerfile#L134

jameslamb (approved):
Thanks for this, I think it's totally fine to add to this image. Really appreciate the testing you did in `rmm`, that + my read of https://github.com/pyenv/pyenv/blob/master/COMMANDS.md makes me pretty confident in this change.

I put up one small suggestion... I'm going to apply it in the UI and then merge this. I strongly suspect you won't object, and it'd be easy enough to revert.

PR #167, opened 2024-08-08
[WIP] Merge miniforge-cuda into ci-imgs

hcho3:
Closes https://github.com/rapidsai/build-planning/issues/48

### DIFF
--- a/.github/workflows/build-image.yaml
+++ b/.github/workflows/build-image.yaml
@@ -21,6 +21,13 @@ on:
       IMAGE_NAME:
         required: true
         type: string
+      DOCKERFILE:
+        required: true
+        type: string
+      DOCKER_TARGET:
+        required: false
+        type: string
+
 
 jobs:
   run:
@@ -62,9 +69,10 @@ jobs:
         uses: docker/build-push-action@v6
         with:
           context: context
-          file: ${{ inputs.IMAGE_REPO }}.Dockerfile
+          file: ${{ inputs.DOCKERFILE }}
           push: true
           pull: true
           build-args: |
             ${{ steps.generate-build-args.outputs.ARGS }}
+          target: ${{ inputs.DOCKER_TARGET }}
           tags: ${{ inputs.IMAGE_NAME }}-${{ matrix.ARCH }}
--- a/.github/workflows/build-and-publish-images.yaml
+++ b/.github/workflows/build-and-publish-images.yaml
@@ -40,6 +40,8 @@ jobs:
       PYTHON_VER: ${{ matrix.PYTHON_VER }}
       IMAGE_REPO: ${{ matrix.IMAGE_REPO }}
       IMAGE_NAME: ${{ matrix.IMAGE_NAME }}
+      DOCKERFILE: ${{ matrix.DOCKERFILE }}
+      DOCKER_TARGET: ${{ matrix.DOCKER_TARGET }}
   build-multiarch-manifest:
     name: manifest (${{ matrix.CUDA_VER }}, ${{ matrix.PYTHON_VER }}, ${{ matrix.LINUX_VER }}, ${{ matrix.IMAGE_REPO }})
     needs: [docker, compute-matrix]
--- a/ci-conda.Dockerfile
+++ b/ci-conda.Dockerfile
@@ -1,21 +1,88 @@
 ARG CUDA_VER=notset
 ARG LINUX_VER=notset
 ARG PYTHON_VER=notset
-ARG YQ_VER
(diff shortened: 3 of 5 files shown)


latest.yaml
```diff
@@ -1,4 +1,8 @@
 # Define the values used for the "latest" tag
+base:
```
bdice, line 2: Can we rename `base` to something like `miniforge-cuda`? This image is not a base for the wheel CI images.

Likewise for `conda` -> `ci-conda`. I'd like to use key names that match `${IMAGE_REPO//-/_}` for clarity.

.github/workflows/build-image.yaml
```diff
@@ -21,6 +21,13 @@ on:
       IMAGE_NAME:
         required: true
         type: string
+      DOCKERFILE:
+        required: true
+        type: string
+      DOCKER_TARGET:
+        required: false
+        type: string
+
```
jakirkham, line 31: nit: Think we can drop this extra blank line

```suggestion

```

ci-wheel.Dockerfile
```diff
@@ -4,7 +4,7 @@ ARG LINUX_VER=notset
 ARG BASE_IMAGE=nvcr.io/nvidia/cuda:${CUDA_VER}-devel-${LINUX_VER}
 ARG AWS_CLI_VER
```
jakirkham, line 5: Should we add `notset` here too? Think this may help with the GHA lints

```suggestion
ARG AWS_CLI_VER=notset
```

citestwheel.Dockerfile
```diff
@@ -4,7 +4,7 @@ ARG LINUX_VER=notset
 ARG BASE_IMAGE=nvcr.io/nvidia/cuda:${CUDA_VER}-devel-${LINUX_VER}
 ARG AWS_CLI_VER
```
jakirkham, line 5: Same suggestion as above

```suggestion
ARG AWS_CLI_VER=notset
```

jakirkham (commented):
Thanks Hyunsu! 🙏

Had a few minor suggestions below

jameslamb (commented):
Thanks very much! Overall, this looks good to me I like the approach of sharing a single Dockerfile for `ci-conda` and `miniforge-cuda`.

Some things I checked, that are working well:

* the matrix in `matrix.yaml` here as of this PR exactly matches https://github.com/rapidsai/miniforge-cuda/blob/main/matrix.yaml
* the `Dockerfile` changes here accurately capture the content of https://github.com/rapidsai/miniforge-cuda/blob/main/Dockerfile
* the patterns for tags match what's been published at https://hub.docker.com/r/rapidsai/miniforge-cuda/tags (thanks especially for catching the `-base-` for `miniforge-cuda`, I might have missed that)

However, I see 2 things that should be changed:

### Ensure intermediate images are getting cleaned up in `rapidsai/staging:` on DockerHub

I think maybe something is broken with the jobs that delete temporary images?

I still see lots of `*-167-*` images at https://hub.docker.com/r/rapidsai/staging/tags?page=&page_size=&ordering=&name=-167-. That `167` corresponds to the PR number here on GitHub.

In theory, I think those should be getting removed by this:

https://github.com/rapidsai/ci-imgs/blob/358b5d1317945de4fe0e7da59274e1ab652f76ae/.github/workflows/build-and-publish-images.yaml#L67

Could you please look into that?

### This PR does not finish the work

This PR shouldn't say it "closes" the work to move `miniforge-cuda` into this repo. Once this is merged and a few days have gone by with no issues, there are a few other things I think need to be done to consider this work complete. I just put those up in a task list at https://github.com/rapidsai/build-planning/issues/48.

latest.yaml
```diff
@@ -1,9 +1,17 @@
 # Define the values used for the "latest" tag
-conda:
+miniforge-cuda:
   CUDA_VER: "12.2.2"
```
bdice, line 3: We need to update all these to 12.5. But that can be a follow-up PR.

bdice (approved):
This looks fine to me, aside from @jameslamb's comments on intermediate image cleanup and follow-up tasks.

bdice:
> ### Ensure intermediate images are getting cleaned up in rapidsai/staging: on DockerHub
>
> I think maybe something is broken with the jobs that delete temporary images?

#172 is handling this in the same way as https://github.com/rapidsai/docker/pull/709.

jameslamb (approved):
Changes in this PR look good and we should merge this. But please do see the other follow-up items I mentioned before we consider https://github.com/rapidsai/build-planning/issues/48 complete.

PR #168, opened 2024-08-15
Build and install aws-sdk-cpp in wheel images

KyleFromNVIDIA:
Contributes to https://github.com/rapidsai/kvikio/pull/426

### DIFF
--- a/ci-wheel.Dockerfile
+++ b/ci-wheel.Dockerfile
@@ -128,7 +128,7 @@ cd aws-sdk-cpp
 cmake \
   -S . \
   -B build \
-  --install-prefix /usr/local \
+  -DCMAKE_INSTALL_PREFIX=/usr/local \
   -DCMAKE_BUILD_TYPE=Release \
   -DBUILD_ONLY=s3 \
   -DBUILD_SHARED_LIBS=OFF \
--- a/ci-wheel.Dockerfile
+++ b/ci-wheel.Dockerfile
@@ -132,7 +132,8 @@ cmake \
   -DCMAKE_BUILD_TYPE=Release \
   -DBUILD_ONLY=s3 \
   -DBUILD_SHARED_LIBS=OFF \
-  -DENABLE_TESTING=OFF
+  -DENABLE_TESTING=OFF \
+  -DENABLE_UNITY_BUILD=ON
 cmake --build build/
 cmake --install build/
 popd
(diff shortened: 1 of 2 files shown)


ci-wheel.Dockerfile
```diff
@@ -119,6 +119,25 @@ mv gh_*/bin/gh /usr/local/bin
...
+git clone --recurse-submodules -b ${AWS_SDK_CPP_VER} https://github.com/aws/aws-sdk-cpp.git
+cd aws-sdk-cpp
+cmake \
+  -S . \
+  -B build \
+  -DCMAKE_INSTALL_PREFIX=/usr/local \
+  -DCMAKE_BUILD_TYPE=Release \
+  -DBUILD_ONLY=s3 \
+  -DBUILD_SHARED_LIBS=OFF \
+  -DENABLE_TESTING=OFF
```
jameslamb, line 135:
```suggestion
  -DENABLE_TESTING=OFF \
  -DENABLE_UNITY_BUILD=ON
```

What do you think about enabling this option?

From https://github.com/aws/aws-sdk-cpp/blob/af521afd75449b7617cb27e8fff6adf67e35d613/CMakeLists.txt#L31

> _"If enabled, the SDK will be built using a single unified .cpp file for each service library.  Reduces the size of static library binaries on Windows and Linux"_

That seems useful for our purposes.

jameslamb (approved):
Given the choice to build this from source in these images, the implementation looks good to me. Thanks for also adding it to the `renovate` config!

I left one small suggestion for your consideration. Do what you want with it, and I don't need to review again.

ci-wheel.Dockerfile
KyleFromNVIDIA: Good idea. I've turned on unity build - we'll see how well it works.

vyasr (approved):
Looks fine, although I didn't review the full set of available build options for aws-sdk-cpp.

ci-wheel.Dockerfile
```diff
@@ -119,6 +119,26 @@ mv gh_*/bin/gh /usr/local/bin
...
+  -B build \
+  -DCMAKE_INSTALL_PREFIX=/usr/local \
+  -DCMAKE_BUILD_TYPE=Release \
+  -DBUILD_ONLY=s3 \
+  -DBUILD_SHARED_LIBS=OFF \
+  -DENABLE_TESTING=OFF \
+  -DENABLE_UNITY_BUILD=ON
+cmake --build build/
+cmake --install build/
+popd
```
trxcllnt, line 139: Should `rm -rf /tmp/aws-sdk-cpp` here so it isn't included in this layer

PR #169, opened 2024-08-15
stop automatically recommending sccache updates

jameslamb:
Replaces #144 

Based on an offline conversation with @trxcllnt . `sccache` updates have such a high potential to disrupt development in RAPIDS, I think we should only update to a new version manually and based on some specific need (getting a known new feature, upgrading past a known bug, etc.).

This proposes that change.

### DIFF
--- a/versions.yaml
+++ b/versions.yaml
@@ -1,4 +1,6 @@
-# renovate: datasource=github-releases depName=mozilla/sccache
+# 'sccache' is intentionally not auto-updated by renovate. Such updates
+# have the potential to significantly disrupt RAPIDS development, so we
+# prefer sticking with a stable version until a specific reason to update arises.
 SCCACHE_VER: 0.7.7
 # renovate: datasource=github-releases depName=cli/cli
 GH_CLI_VER: 2.54.0


AyodeAwe (approved):

KyleFromNVIDIA (approved):
Let's make sure to never use the versions of `sccache` that prompted the Great Cache Dump from earlier this year due to https://github.com/mozilla/sccache/issues/2038... hopefully this change will prevent such disasters in the future

jakirkham:
How do we plan to stay on top of updates?

jameslamb:
> How do we plan to stay on top of updates?

Not touching this unless and until there's something specific we want in a specific new version.

This is what we've already been doing in practice. There have been newer versions available since May 31st (#144) that we've been choosing not to upgrade to.

This PR just turns that from a convention into something explicit, to make it clear that updating this is something that requires great care and that shouldn't be done as routine maintenance.

jameslamb:
Based on approvals (including 👍🏻 s) I think we have agreement here. Going to merge this. Thanks for the suggestion @trxcllnt !

PR #170, opened 2024-08-15
Fix GHA warnings in diffs

jakirkham:
Our `*.Dockerfile`s are getting flagged for warnings by GHA. However each time a job runs it gets flagged. So we get dozens of comments in the diffs, which makes reviewing PRs here difficult

To fix this issue this cleans up the common warnings we have seen

We may also want to separately consider only raising these warnings once so as to keep PR reviews manageable

Edit: For clarity these are the GHA warnings that I see currently. It takes quite a bit of scrolling to get through them all

<details>
Screenshot 2024-08-15 at 2 39 48 PM
</details>


bdice (approved):
This is also handled in #167. We can merge either PR to move forward.

jakirkham:
Thanks Bradley! 🙏

Honestly am having a hard time reviewing PRs because of these warnings. So wanted to speed up resolution. Hence addressing in this PR

PR #177, opened 2024-08-29
Add Rocky Linux 8 citestwheel images.

bdice:
This adds Rocky Linux 8 citestwheel images. We would like to test wheels on Rocky Linux 8 because it has glibc 2.28, which should give us more confidence about the functionality of our manylinux 2.28 wheels.


jameslamb (approved):

bdice:
I’m going to merge this and open a few test PRs to see how things go. It may take some iterations to get these kmages working well.

PR #183, opened 2024-09-03
clarify 'docker' job names in GitHub UI

jameslamb:
On #179, I saw that some subset of image builds was deterministically failing. It was difficult to tell which subset, because the names for the main image-building job are long and repetitive.

This shortens those names, to hopefully make such comparisons easier.

Before:

image

After:

image


ajschmidt8 (approved):
nice! looks good to me.

bdice (approved):

PR #186, opened 2024-09-03
ci-conda: update environment earlier

### LINKED ISSUE #185: ci-conda builds failing with GLIBCXX errors
## Description

48 of the `ci-conda` image builds jobs are deterministically failing with GLIBC errors like this:

> ImportError: /lib/x86_64-linux-gnu/libstdc++.so.6: version `GLIBCXX_3.4.29' not found (required by /opt/conda/lib/python3.11/site-packages/libmambapy/bindings.cpython-311-x86_64-linux-gnu.so)

## Reproducible Example

Observed this on multiple unrelated PRs, e.g. #179 and #183.

Example build link: https://github.com/rapidsai/ci-imgs/actions/runs/10686297019/job/29621203550?pr=183

## Notes

N/A

jameslamb:
Fixes #185

https://github.com/rapidsai/ci-imgs/issues/185#issuecomment-2327478322 has MUCH more detail, but in short:

* the way we build `ci-conda` images results in symlinks from `libdstc++.so` to the fully-qualified `libstdc++.so.6.0.33` (or similar) being destroyed
* that results in the system `libdstc++` being used instead of conda's
* that results in issues about missing GLIBCXX symbols (in the case where the system one is older than the one used at build time for libraries conda wants to load)

~**This PR is a hack to unblock CI**... by manually recreating those symlinks where necessary.~

Updated this to @vyasr 's suggestion from https://github.com/rapidsai/ci-imgs/issues/185#issuecomment-2327725065

### DIFF
--- a/ci-conda.Dockerfile
+++ b/ci-conda.Dockerfile
@@ -38,6 +38,19 @@ if [[ "$LINUX_VER" == "rockylinux"* ]]; then
 fi
 find /opt/conda -follow -type f -name '*.a' -delete
 find /opt/conda -follow -type f -name '*.pyc' -delete
+# Recreate missing libstdc++ symlinks.
+# This should be removed when it is fixed upstream.
+# ref: https://github.com/rapidsai/ci-imgs/issues/185
+if [[ ! -e "/opt/conda/lib/libstdc++.so.6" ]]; then
+    ln -sf \
+        "$(ls /opt/conda/lib/libstdc++.so.6.* | head -1)" \
+        /opt/conda/lib/libstdc++.so.6
+fi
+if [[ ! -e "/opt/conda/lib/libstdc++.so" ]]; then
+    ln -sf \
+        "$(ls /opt/conda/lib/libstdc++.so.6.* | head -1)" \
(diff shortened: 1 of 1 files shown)


hcho3 (approved):

vyasr (changes requested):
I'd like to propose an alternate, hopefully more robust solution to the problem. [I posted it on the issue](https://github.com/rapidsai/ci-imgs/issues/185#issuecomment-2327725065).

ci-conda.Dockerfile
```diff
@@ -29,6 +29,9 @@ RUN chmod g+ws /opt/conda
 RUN <<EOF
 # Ensure new files/dirs have group write permissions
 umask 002
+# update everything before other environment changes, to ensure mixing
+# and older conda with newer packages still works well
```
vyasr, line 33:
```suggestion
# an older conda with newer packages still works well
```

vyasr (approved):
One typo otherwise LGTM.

ci-conda.Dockerfile
jameslamb: 🙈 thank you

PR #188, opened 2024-09-09
update 'latest' tags to Python 3.12 images

jameslamb:
Contributes to https://github.com/rapidsai/build-planning/issues/40

Ensures that the images tagged `:latest` here use Python 3.12. Also suggests some small changes to the project's docs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,12 +12,12 @@ The `latest` image tags are controlled by the values in `latest.yaml`.
 
 ## Building the dockerfiles locally
 
-To easily build the dockerfiles locally, you may use the following snippets:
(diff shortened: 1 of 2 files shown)


KyleFromNVIDIA (approved):

bdice (approved):
Approving. Note that if we have to do another hotfix of 24.08 we would need to revert this. I support merging this, now that the polars release is complete.

jameslamb:
> Note that if we have to do another hotfix of 24.08 we would need to revert this.

Ah ok yep, that's a good point. I also agree this could be merged right now.

PR #189, opened 2024-09-24
Add packaging lib to ci-conda image

AyodeAwe:
Installs the [packaging lib](https://pypi.org/project/packaging/#history), which is used in `rapids-pip-wheel-version` gha tool.

Related:
- https://github.com/rapidsai/gha-tools/pull/117

### DIFF
--- a/ci-conda.Dockerfile
+++ b/ci-conda.Dockerfile
@@ -231,6 +231,9 @@ RUN <<EOF
 pip install codecov-cli==${CODECOV_VER}
 EOF
 
+# Install packaging lib
+RUN pip install packaging
+
 RUN /opt/conda/bin/git config --system --add safe.directory '*'
 
 COPY --from=yq /usr/bin/yq /usr/local/bin/yq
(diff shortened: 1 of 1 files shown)


ci-conda.Dockerfile
```diff
@@ -231,6 +231,9 @@ RUN <<EOF
 pip install codecov-cli==${CODECOV_VER}
 EOF
 
+# Install packaging lib
+RUN pip install packaging
```
jameslamb, line 235: Since this is for the `ci-conda` image, I think it'd be better to use `conda` to install this. That tends to lead to less surprises than mixing `pip` and `conda` installed things in the same environment.

Instead of a new layer, could you just add this to this block?

https://github.com/rapidsai/ci-imgs/blob/0b910783e9ba8dd7e17b123a972f46875883c2e0/ci-conda.Dockerfile#L196-L207

I just double-checked, and it's literally called `packaging` on `conda-forge` too (https://github.com/conda-forge/packaging-feedstock).

jameslamb (changes requested):
I think we should use conda to install this, and avoid adding another layer to the image.

jameslamb (approved):

PR #190, opened 2024-09-25
use micromamba to create the python envs

msarahan:
This is something I worked on a while back, then got sidetracked. I'm putting it up as a PR for visibility, not for review at the moment.

Related to https://github.com/rapidsai/ci-imgs/issues/185

### DIFF
--- a/ci-conda.Dockerfile
+++ b/ci-conda.Dockerfile
@@ -1,19 +1,18 @@
-ARG CUDA_VER=notset
-ARG LINUX_VER=notset
-ARG PYTHON_VER=notset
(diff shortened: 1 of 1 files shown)


vyasr:
Please also link https://github.com/rapidsai/build-planning/issues/50 (and mark this as closing that if this is the only PR needed to make that happen, although I'm not sure that it is).

bdice:
@msarahan Is this still worth pursuing? Or should we close it?

vyasr:
I think it's still worth doing, but low priority. See https://github.com/rapidsai/build-planning/issues/50.

PR #184, opened 2024-09-03
Add Python 3.9 back temporarily for a `24.08` hotfix

raydouglass:
We're planning one final hotfix for `24.08` which supports Python 3.9, so add it back here out an abundance of caution. Without this, the Python 3.9 builds will use outdated images.

This can undone in ~2 weeks.

PR #173, opened 2024-08-27
Drop Python 3.9, clean up CentOS references, update latest.yaml

bdice:
This PR does a few update/cleanup tasks:

- Drop Python 3.9.
- Cleanup remaining centos references.
- Update latest images.

PR #194, opened 2024-10-02
Drop Python 3.9, simplify ci-wheel matrix (only oldest glibc and latest CUDA of each major version).

bdice:
This PR drops Python 3.9 (see #173, #184) now that we no longer need it for 24.10 releases and do not have any further hotfixes planned for 24.08.

### DIFF
--- a/matrix.yaml
+++ b/matrix.yaml
@@ -5,7 +5,6 @@ CUDA_VER:
   - "12.2.2"
   - "12.5.1"
 PYTHON_VER:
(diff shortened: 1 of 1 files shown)


raydouglass (approved):

jameslamb (approved):
thank you! I agree this is safe now.

bdice:
Awesome, thanks @raydouglass / @jameslamb. CI passed so I'll merge this.

PR #198, opened 2024-10-16
Set default value of RAPIDS_DEPENDENCIES="latest"

bdice:
Recently, we added the ability to use `"oldest"` or `"latest"` dependencies in RAPIDS CI workflows. https://github.com/rapidsai/shared-workflows/pull/228

However, this has hurt our ability to reproduce CI workflows locally, because `RAPIDS_DEPENDENCIES` is not set by default in our CI containers. This leads to errors like the following when running CI images locally:

```
./ci/test_python_common.sh: line 17: RAPIDS_DEPENDENCIES: unbound variable
```

This PR adds a default value of `RAPIDS_DEPENDENCIES="latest"` to aid in reproducing CI workflows. Our workflows can still set it to a non-default value to test `"oldest"` dependencies.


jameslamb (approved):
yes totally, support it

PR #207, opened 2024-11-19
Add Python 3.13

bdice:
This PR adds Python 3.13 CI images for `ci-wheel` and `citestwheel`.

For now we are blocked by https://github.com/conda-forge/conda-feedstock/issues/246 on adding Python 3.13 support to our conda CI images. However, I went ahead and did some refactoring that will be needed for conda support once that is available.


jakirkham:
This is now failing as `boa` requires an older version of `conda`, which is incompatible with Python 3.13

Sounds like we should prioritize dropping `boa` to move Python 3.13 forward: https://github.com/rapidsai/build-planning/issues/149

bdice:
I wrote up an automation script that we can use to convert all RAPIDS repositories from `conda mambabuild` to `conda build`, which would allow us to drop `boa`. However, we are also working through a migration to `rattler-build` which will effectively drop `boa`.

Even if we dropped `boa`, we would still be blocked from migrating to Python 3.13 because cupy doesn't support Python 3.13 yet. https://github.com/cupy/cupy/issues/8651

In the interest of keeping our total work at a minimum: if cupy and other RAPIDS dependencies become available for Python 3.13 but still haven't finished the `rattler-build` upgrade, then we can migrate all remaining RAPIDS repositories off of `boa` to unblock the Python 3.13 upgrade.

ci-conda.Dockerfile
```diff
@@ -212,9 +221,18 @@ RUN cat /tmp/condarc.tmpl | envsubst | tee /opt/conda/.condarc; \
...
+PYTHON_MINOR_VERSION=${PYTHON_VERSION#*.}
+PYTHON_UPPER_BOUND="${PYTHON_MAJOR_VERSION}.$((PYTHON_MINOR_VERSION+1)).0a0"
+PYTHON_MINOR_PADDED=$(printf "%02d" "$PYTHON_MINOR_VERSION")
+PYTHON_VERSION_PADDED="${PYTHON_MAJOR_VERSION}.${PYTHON_MINOR_PADDED}"
+if [[ "$PYTHON_VERSION_PADDED" > "3.12" ]]; then
+    PYTHON_ABI_TAG="cp${PYTHON_MAJOR_VERSION}${PYTHON_MINOR_VERSION}"
+else
+    PYTHON_ABI_TAG="cpython"
+fi
 rapids-mamba-retry install -y \
```
bdice, line 234:
```suggestion

# rust is needed for source builds of codecov-cli -- binaries are not available for Python 3.13 yet.
rapids-mamba-retry install -y \
```

bdice:
This should be unblocked now.

jameslamb (approved):
Double-checked a couple build logs, saw the things I expected... the right build-string filters to say "only give us CPython", the right Python versions installed, and no new warnings to worry about.

I'm ok with duplicating the `PYTHON_*_VERSION` logic in the Dockerfile instead of doing that in `compute-build-args.sh`, think that's fine.

PR #214, opened 2024-12-02
ci-conda: change spelling of python dependency

jameslamb:
Proposes changing the spelling of the Python dependency for conda builds here from something like this:

```text
python~=3.11.0=*_cpython
```

to something like this:

```text
python>=3.11,<3.12.0a0=*_cpython
```

To work around an issue with parsing `~=` in recent versions of `mamba`.

References:

* https://github.com/rapidsai/ci-imgs/pull/209#issuecomment-2511996860
* https://github.com/conda/conda-libmamba-solver/issues/581
* https://github.com/mamba-org/mamba/issues/3647


jameslamb:
It looks to me like this is working! For example, in a successful `miniforge-cuda` Python 3.11 build, I see the following:

```text
  PYTHON_VER=3.10 
  PYTHON_VER_UPPER_BOUND=3.11.0a0 
...
#10 26.22   added / updated specs:
#10 26.22     - python[version='>=3.10,<3.11.0a0',build=*_cpython]
...
#10 26.22   python                          3.12.7-h5d932e8_0_cpython --> 3.10.15-hbf90c55_2_cpython 
#10 26.22   python_abi                                   3.12-5_cp312 --> 3.10-5_cp310 
```

([build link](https://github.com/rapidsai/ci-imgs/actions/runs/12125650004/job/33806923867?pr=214#step:8:372))

bdice (approved):
Thanks for the fix. For awareness, in the past we just used pinnings like `python=3.10`. See here for historical context: https://github.com/rapidsai/miniforge-cuda/pull/52/files

jameslamb:
> in the past we just used pinnings like `python=3.10`

That would definitely be simpler, and as far as I know should still achieve what we want.. @jakirkham what was the reason for switching from `=` to `~=` in https://github.com/rapidsai/miniforge-cuda/pull/52?

Looks like that was also carried over to this repo #167, but I don't see any discussion about it there. (cc @hcho3 in case you remember)

I'll merge this right now to unblock CI, but would be happy to come back and switch this to `=` (+ revert this logic about setting an upper bound) if we find out that that would work just as well.

bdice:
Also in #207 I had to add some logic to use a different build tag. It would be easy to compute the upper bound there rather than supplying it as an input. We can discuss on that PR.

PR #215, opened 2024-12-03
add 'wheel' to ci-wheel image

### LINKED ISSUE #138: Update for RAPIDS C++ wheels
Once https://github.com/rapidsai/build-planning/issues/33 is complete, we will have a new set of wheels that change some of the requirements of our CI images. This issue is to keep track of such changes as we discover them so that we can make them all at the end.

- [x] We will no longer need to install UCX into the images since UCX will be available as a wheel.
  - #154
- [ ] We will need to install the wheel Python package in addition to auditwheel because for the header-only C++ libraries the wheels won't contain any binaries, so auditwheel will be a no-op (`auditwheel repair` will not retag wheels that have no binaries to repair).
- #215

jameslamb:
Closes #138

Proposes installing `wheel` in the `ci-wheel` images, to remove the need to install it at build time on every CI run for header-only C++ wheel builds, like `librmm` does:

```shell
python -m pip install wheel
python -m wheel tags --platform any dist/* --remove
```

([rmm/ci/build_wheel.sh](https://github.com/rapidsai/rmm/blob/d4066fa611c803430c9bc5dbe8e243f89bb9a25c/ci/build_wheel_cpp.sh#L23-L24))

### DIFF
--- a/ci-wheel.Dockerfile
+++ b/ci-wheel.Dockerfile
@@ -139,7 +139,8 @@ python -m pip install \
   patchelf \
   'pydistcheck==0.8.*' \
   'rapids-dependency-file-generator==1.*' \
-  twine
+  twine \
+  wheel
 pyenv rehash
 EOF
 


bdice (approved):
Seems like a reasonable addition to me. 🛞

Issue #162, opened 2024-07-15
GHA jobs fail because of too many requests


jakirkham:
Have started seeing GHA jobs fail because of too many request (for [example]( https://github.com/rapidsai/ci-imgs/actions/runs/9947106104/job/27479174316#step:8:1276 ):

```
#26 ERROR: toomanyrequests: Too Many Requests (HAP429).

------
 > pushing rapidsai/ci-conda:cuda11.8.0-rockylinux8-py3.10-amd64 with docker:
------
```

This is the same as issue ( https://github.com/rapidsai/miniforge-cuda/issues/72 ), which already contains suggestions on how this could be resolved

jameslamb:
I just observed the same thing on https://github.com/rapidsai/ci-imgs/actions/runs/11614334107/job/32342353447

43 of 270 jobs failed because they hit this rate limit. All succeeded on a rebuild. Just posting since this is an old issue, to note that this is still a problem we hit sometimes.

jameslamb:
I jsut merged #216, which limits this repo to running 150 jobs at a time within a workflow run. Let's see if that helps.

vyasr:
@jameslamb what data do we want to see before we close this issue?

jameslamb:
🤷🏻 I don't know, there are so many factors that affect whether or not we hit this.

Let's say when the next round of `renovate` PRs go up on January 1st, if we don't see any rate-limiting errors, we close this.

jameslamb:
In #219 we reduced this further to 50 parallel jobs.

Saw multiple runs on different days where most jobs succeeded and **0 had rate-limiting issues**. ref: https://github.com/rapidsai/ci-imgs/pull/219#issuecomment-2596481942

If the job triggered by the merge of #219 also succeeds without any rate-limiting issues, I think we should close this. Build to watch: https://github.com/rapidsai/ci-imgs/actions/runs/12816222911

jameslamb:
> If the job triggered by the merge of https://github.com/rapidsai/ci-imgs/pull/219 also succeeds without any rate-limiting issues, I think we should close this. Build to watch: https://github.com/rapidsai/ci-imgs/actions/runs/12816222911

Pretty similar to what we saw on #219... 265 out of the 270 build jobs succeeded, and that took about 27 minutes.

Importantly, **none of the failures were related to rate-limiting**

* conda/mamba errors ("Unexpected error 9 on netlink descriptor 9.")
[...]

PR #216, opened 2024-12-03
limit to 150 concurrent jobs per workflow

jameslamb:
Contributes to #162

Since I started paying close attention, I've observed that most CI runs here have at least 10 failures on their first run, with errors indicating we hit dockerhub rate limits:

```text
ERROR: toomanyrequests: Too Many Requests (HAP429).
```

This proposes using @ajschmidt8 's recommendation from https://github.com/rapidsai/miniforge-cuda/issues/72#issuecomment-2229460639 to limit the number of concurrent jobs.

### Benefits of this change

Reduces the impact of this repo on total CPU runner availability for projects using NVIDIA runners.

Reduces the likelihood that a human will have to retrigger a build here (which costs time and money, and is easy to miss on branch builds after merges).

### Why set the limit to 150?

This is not an exact science haha. I'm just looking for a number that meets these constraints:

* fewer concurrent jobs than this repo currently uses
* does not increase CI times *too much*

Some relevant information:

* current number of build jobs per workflow run: 270
* worst-case (uncached) DockerHub image pulls per workflow run: 450
  - *`ci-conda` / `miniforge-cuda`:*
    - *pulls: 4 (`nvidia/cuda`, `condaforge/miniforge3`, `mikefarah/yq`, `amazon/aws-cli`)*
    - *builds: 90*
    - *total: 360*
  - *`ci-wheel`:*
    - *pulls: 1 (`amazon/aws-cli`... base image is from NVCR)*
    - *builds: 12*
    - *total: 12*
  - *`citestwheel`:*
    - *pulls: 1 (`amazon/aws-cli` ... base image is from NVCR)*
    - *builds: 78*
    - *total: 78*

So assuming full availability of `linux-{aarch64,amd64}-cpu` runners, and if all build jobs take roughly the same amount of time, changing from "unlimited" to 150 might mean roughly 1.8x the end-to-end time for a CI run here... from around 11 minutes to maybe 20 minutes.

I have no idea what the exact limit is from DockerHub. From https://docs.docker.com/docker-hub/download-rate-limit/#other-limits:

> *Docker Hub also has an overall rate limit... This limit applies to all requests to Hub properties including web pages, APIs, and image pulls. The limit is applied per-IP, and... the limit changes over time...it's in the order of thousands of requests per minute. ... [and] applies to all users equally regardless of account level.*

> *The "overall limit" returns a simple 429 Too Many Requests response. The pull limit returns a longer error message that includes a link to this page.*

### DIFF
--- a/.github/workflows/build-and-publish-images.yaml
+++ b/.github/workflows/build-and-publish-images.yaml
@@ -32,6 +32,7 @@ jobs:
     name: ${{ matrix.IMAGE_REPO }}
     needs: compute-matrix
     strategy:
+      max-parallel: 150
       matrix: ${{ fromJSON(needs.compute-matrix.outputs.MATRIX) }}
       fail-fast: false
     secrets: inherit
--- a/.github/workflows/build-image.yaml
+++ b/.github/workflows/build-image.yaml
@@ -32,6 +32,7 @@ jobs:
   run:
     name: build (${{ matrix.CUDA_VER }}, ${{ matrix.PYTHON_VER }}, ${{ matrix.LINUX_VER }}, ${{ matrix.ARCH }})
     strategy:
+      max-parallel: 150
       matrix:
         ARCH: ${{ fromJSON(inputs.ARCHES) }}
         CUDA_VER: ["${{ inputs.CUDA_VER }}"]


bdice (approved):
If this works on the first run, we can merge it. I agree it's not an exact science, and 150 seems like a reasonable value.

PR #217, opened 2024-12-09
Enable 'quiet: true' in condarc

bdice:
See https://github.com/rapidsai/build-planning/issues/126.

We are getting lots and lots of noisy log lines with no content, because progress bars are being included in our CI log lines. This sets `quiet: true` for conda to help with the problem.

PR #220, opened 2024-12-23
Revert "Enable 'quiet: true' in condarc"

bdice:
Reverts rapidsai/ci-imgs#217.

It seems like we have lost some useful information from conda, and Python exception names are no longer being printed. This makes rapids-conda-retry ineffective. @pentschev and I suspect the change to quiet mode may be the root cause. We would like to try reverting it to see if this changes.

We may also need to do something to make conda stop printing all the extra newlines in progress bars that led to #217 in the first place…

### DIFF
--- a/context/condarc.tmpl
+++ b/context/condarc.tmpl
@@ -13,4 +13,3 @@ conda-build:
   output_folder: $RAPIDS_CONDA_BLD_OUTPUT_DIR
 number_channel_notices: 0
 always_yes: true
-quiet: true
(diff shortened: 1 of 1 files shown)


pentschev (approved):
Thanks @bdice !

jakirkham:
Could you please share a bit more what information is lost?

Some examples would be instructive

bdice:
The exact problem is that when we enable quiet mode, Python exception names are no longer being printed. This makes rapids-conda-retry ineffective, as it depends on parsing logs for exception names to decide whether to retry.

bdice:
Ah, there is a better continuation of this discussion in #228. Let’s continue there.

jakirkham:
Please feel free to share some examples in that thread. Links to logs are ok. Just need something with more detail

Issue #227, opened 2025-01-13
Remove pytorch channel


bdice:
Following https://github.com/pytorch/pytorch/issues/138506, RAPIDS has been migrating all packages to use the `conda-forge` distribution of PyTorch instead of the `pytorch` channel.

We should try to remove the `pytorch` channel from the CI images' default list of channels:

https://github.com/rapidsai/ci-imgs/blob/8312f41003e8633a3117feca3fd90778f6d5a41e/context/condarc.tmpl#L6

Validation:
- cudf nightlies use `conda-forge` pytorch packages for [CUDA 12](https://github.com/rapidsai/cudf/actions/runs/12741891101/job/35528673775#step:8:4554) and skip those tests for CUDA 11
- cugraph-gnn uses `conda-forge` pytorch packages for [CUDA 11.8](https://github.com/rapidsai/cugraph-gnn/actions/runs/12722516249/job/35466625866#step:8:6158) and [CUDA 12](https://github.com/rapidsai/cugraph-gnn/actions/runs/12722516249/job/35466626069#step:8:5736)
  - There appear to be dependencies on `captum` (CUDA 11 and 12) and `pytorch-cuda` (only CUDA 11.8) from the PyTorch channel that need to be checked on and removed?

bdice:
I asked about captum here: https://github.com/conda-forge/captum-feedstock/pull/6#issuecomment-2587738287
I think that will be fixed by an easy conda-forge feedstock update.

As for pytorch-cuda, I suspect that is coming from one of the GNN dependencies and will probably disappear once we drop CUDA 11 -- it doesn't show up for CUDA 12 builds currently.

bdice:
I think that `captum` should be fixed now that we have https://github.com/conda-forge/captum-feedstock/pull/7 updating the version to the same version on the `pytorch` channel. However, it may fall behind due to channel priority. I wonder if we can explore moving `conda-forge` _above_ `pytorch` in the short term, and then remove it at a later time?

bdice:
In fact, `cugraph-gnn` doesn't even list `pytorch` in its channel list: https://github.com/rapidsai/cugraph-gnn/blob/87455cfedcc6721f24c783ba555af14a9a180624/dependencies.yaml#L185-L190

[...]

PR #229, opened 2025-01-16
Remove pytorch channel.

bdice:
Closes https://github.com/rapidsai/ci-imgs/issues/227.


jameslamb (approved):

bdice:
I'm going to wait to merge this until after the GCC 13 migration. I want to get that across the line before doing anything potentially disruptive. I do not expect this to be breaking, but just want to be on the safe side.

PR #231, opened 2025-01-23
Update to CUDA 12.8.0.

bdice:
Adds CUDA 12.8.0 CI images.

This depends on `nvidia/cuda` images being available.


bdice:
Hopefully we can file a follow-up PR to drop one of `12.2.2` or `12.5.1` from our CI images. Perhaps we can drop 12.2.2 and keep 12.5.1 as our "middle" CUDA 12 version?

jameslamb:
> Perhaps we can drop 12.2.2 and keep 12.5.1 as our "middle" CUDA 12 version used for testing?

I don't have any specific information to add to help with this decision, but keeping 12.5.1 and dropping 12.2.2 sounds fine to me. I do agree it'd be good to drop one of them from the test matrix, to get denser coverage of 12.8.

bdice:
A bit more context on why I am proposing to keep 12.5 and drop 12.2:

We have a limited amount of CI jobs that we can run. We try to keep dense coverage on the endpoints, like "newest" and "oldest". Historically we have not had a lot of guidance or requirements on which "latest major, middle minor" version to test, and we just used 12.2 because that was our "newest" version until we migrated to 12.5. We don't have enough CI resources to keep both 12.2 and 12.5, so I would favor keeping a more recent version.

jakirkham:
There were a few features that were added in CUDA 12.2 that were not available in previous versions, which motivates testing it

* CUDA 12.2 added cuFile for ARM
  * https://github.com/conda-forge/libcufile-feedstock/pull/9
  * https://github.com/rapidsai/kvikio/pull/405
* cuFile added Stream APIs
  * https://github.com/rapidsai/kvikio/pull/342
* CUDA 12.2 adds HMM
  * https://docs.nvidia.com/cuda/archive/12.2.2/cuda-toolkit-release-notes/index.html#general-cuda
  * Appears this is untested as of yet
    * https://github.com/rapidsai/rmm/issues/1599

Have glanced through the release notes between there and here as well as RAPIDS source. Not seeing anything as compelling for RAPIDS feature-wise as 12.2

So would actually suggest we keep 12.2 since it is a useful lower bound and drop 12.5 in favor of 12.8

vyasr:
I think those are pretty good reasons to prefer 12.2 to 12.5.

bdice:
`nvidia/cuda` images are now live. I will rerun CI.

bdice:
I opened https://github.com/rapidsai/ci-imgs/issues/232 to capture our intentions around dropping CUDA 12.5 images. We should do that before the 25.04 release, to ensure continuity.

vyasr (approved):

jakirkham:
Thanks all! 🙏

PR #235, opened 2025-01-30
Add timeout to image builds.

bdice:
Recently some image builds failed but did not exit for unclear reasons. This resulted in the jobs consuming the full 6 hour time allowed by GitHub Actions.

Examples:
- https://github.com/rapidsai/ci-imgs/actions/runs/13044383798/job/36392385934
- https://github.com/rapidsai/ci-imgs/actions/runs/13044383798/job/36392394499
- https://github.com/rapidsai/ci-imgs/actions/runs/13044383798/job/36392385934

Most jobs take less than 10 minutes, so this PR sets a timeout of 30 minutes to prevent resources from being wasted.


jameslamb (approved):
strong agree, thank you for doing this

PR #243, opened 2025-02-12
added gh cli to citestwheel

VenkateshJaya:
This PR adds GitHub CLI (gh cli) for the `citestwheel`

### DIFF
--- a/citestwheel.Dockerfile
+++ b/citestwheel.Dockerfile
@@ -133,6 +133,7 @@ COPY --from=aws-cli /usr/local/bin/ /usr/local/bin/
 
 # Download and install GH CLI tool
 ARG GH_CLI_VER=notset
+ARG CPU_ARCH=notset
 RUN <<EOF
 set -e
 wget https://github.com/cli/cli/releases/download/v${GH_CLI_VER}/gh_${GH_CLI_VER}_linux_${CPU_ARCH}.tar.gz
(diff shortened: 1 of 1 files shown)


ajschmidt8 (approved):

citestwheel.Dockerfile
```diff
@@ -131,6 +131,17 @@ ENV PATH="/pyenv/versions/${PYTHON_VER}/bin/:$PATH"
 COPY --from=aws-cli /usr/local/aws-cli/ /usr/local/aws-cli/
 COPY --from=aws-cli /usr/local/bin/ /usr/local/bin/
 
+# Download and install GH CLI tool
+ARG GH_CLI_VER=notset
+ARG CPU_ARCH=notset
```
bdice, line 136: @VenkateshJaya Could you file a follow-up to move this `ARG CPU_ARCH=notset` higher up in the file, to match what we do in `ci-wheel`? See https://github.com/rapidsai/ci-imgs/blob/e235facc51dfffe54102a97b83575d2f64e56186/ci-wheel.Dockerfile#L13

Also, could we move this block above `# Install pyenv`? That would make the ordering of the `citestwheel` commands more similar to those in `ci-wheel`.

bdice: I can file a PR to make these changes if someone else would approve it.

VenkateshJaya: I'll do the changes @bdice and open up a new PR

bdice: https://github.com/rapidsai/ci-imgs/pull/245

PR #247, opened 2025-02-28
Update miniforge version, use renovate.

bdice:
Closes #251

This updates our miniforge version and uses renovate to manage versions going forward.


ci-conda.Dockerfile
```diff
@@ -3,6 +3,7 @@ ARG LINUX_VER=notset
 ARG PYTHON_VER=notset
 ARG YQ_VER=notset
 ARG AWS_CLI_VER=notset
+ARG MINIFORGE_VER=notset
```
jameslamb, line 6: Oh sorry, should have caught this.... this `ARG` needs to come down AFTER that first `FROM`.

bdice:
> I support this, but I'm also confused...
> 
> we used to already get changes like this from `renovate`: #208
> 
> Do you understand why we didn't when `24.11.3-0` was tagged?

No, I don’t understand, and I hoped someone else might know.

jameslamb:
Oh hey, we just got one!

https://github.com/rapidsai/ci-imgs/pull/251

So maybe, despite the version indicating November, that `24.11.3-0` was first pushed to DockerHub in mid-February? That'd explain us not getting a new renovate PR until March 1st.

I can't tell, because https://hub.docker.com/r/condaforge/miniforge3/tags they appear to have just very recently overwritten the image at that tag.

Screenshot 2025-03-03 at 9 48 17 AM

ci-conda.Dockerfile
jameslamb: So I'd thought this was just the issue where an `ARG` needs to be restated before statement referencing it, but looks like it actually runs deeper... `docker` / `docker buildx` don't support interpolation with build args in `COPY --from` statements:

* https://github.com/moby/moby/issues/34482
* https://github.com/docker/buildx/issues/2352
* https://stackoverflow.com/a/51983295/3986677

So I just pushed https://github.com/rapidsai/ci-imgs/pull/247/commits/e296ecb5811ba889eca6d4de5a943ca4b3c94ddc to try a workaround... declaring the `miniforge` image in a `FROM` (to alias it to a hard-coded name), then referencing that hard-coded name again in the `COPY --from`. Let's see how that goes.

jameslamb: How I tested this locally:

```shell
docker buildx build \
    --target miniforge-cuda \
    --build-arg AWS_CLI_VER=2.22.2 \
    --build-arg CODECOV_VER=9.1.1 \
    --build-arg CONDA_ARCH=linux64 \
    --build-arg CPU_ARCH=amd64 \
    --build-arg CUDA_VER=11.4.3 \
    --build-arg GH_CLI_VER=2.64.0 \
    --build-arg LINUX_VER=ubuntu20.04 \
    --build-arg MANYLINUX_VER=manylinux_2_31 \
    --build-arg MINIFORGE_VER=24.11.3-0 \
    --build-arg PYTHON_VER_UPPER_BOUND=3.11.0a0 \
    --build-arg REAL_ARCH=x86_64 \
    --build-arg SCCACHE_VER=0.7.7 \
    --build-arg YQ_VER=4.44.6 \
    -f ./ci-conda.Dockerfile \
    ./context
```

jameslamb:
> I'd like to propose that we merge this after the 25.02 patch releases are all done, and close https://github.com/rapidsai/ci-imgs/pull/251.

Agreed! This is passing now, so we can merge whenever 25.02 patch releases are done.

jameslamb:
Merging this now that the 25.02 patch releases are done.

PR #254, opened 2025-03-04
updates: codecov-cli (10.2.0), gh CLI (2.67.0), AWS CLI (2.24.16), yq (4.45.1)

jameslamb:
Closes #250
Closes #252
Closes #253

Consolidates all the other automatic updates for dependencies here.

### DIFF
--- a/versions.yaml
+++ b/versions.yaml
@@ -5,7 +5,7 @@ SCCACHE_VER: 0.7.7
 # renovate: datasource=github-releases depName=cli/cli
 GH_CLI_VER: 2.67.0
 # renovate: datasource=pypi depName=codecov-cli
-CODECOV_VER: 9.1.1
+CODECOV_VER: 10.2.0
 # renovate: datasource=docker depName=mikefarah/yq versioning=docker
 YQ_VER: 4.45.1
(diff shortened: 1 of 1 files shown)


versions.yaml
```diff
@@ -3,12 +3,12 @@
 # prefer sticking with a stable version until a specific reason to update arises.
 SCCACHE_VER: 0.7.7
 # renovate: datasource=github-releases depName=cli/cli
-GH_CLI_VER: 2.64.0
+GH_CLI_VER: 2.67.0
 # renovate: datasource=pypi depName=codecov-cli
 CODECOV_VER: 9.1.1
```
bdice, line 8: I'd like to see if we can update codecov-cli at the same time. (I don't know how Renovate picks which things to update.)
```suggestion
CODECOV_VER: 10.2.0
```

bdice (approved):
Approving, with one suggestion for expansion. Also please merge the upstream so we test these updates with Python 3.13.

versions.yaml
jameslamb: Sure, seems fine to me. I don't understand why we didn't get a PR for this either... there are definitely `codecov-cli` 10.x packages at https://pypi.org/project/codecov-cli/#history, the first of which was released in January.

PR #256, opened 2025-03-18
Force remove package cache and remove rust from CI images

bdice:
This PR improves ci-conda image sizes.

### Purge the package cache

Currently we are shipping a large package cache in `/opt/conda/pkgs` that is not needed.
```
$ docker run -it rapidsai/ci-conda:latest bash -c 'du -sh /opt/conda/*'
...
1.1G    /opt/conda/pkgs
```

The guidance from `conda clean --help` says:
```
-p, --packages        Remove unused packages from writable package caches. WARNING: This does not check for packages installed using symlinks back to the package cache.
...
-f, --force-pkgs-dirs
                      Remove *all* writable package caches. This option is not included with the --all flag. WARNING: This will break environments with packages installed using symlinks back to the package cache.
```

We don't install any packages with symlinks back to the package cache. I verified that removing the `/opt/conda/pkgs` directory does not break any symlinks.

```
# conda clean -f
# cd /opt/conda && find . -xtype l
./x86_64-conda-linux-gnu/lib/libgfortran.so.5
./x86_64-conda-linux-gnu/lib/libgfortran.so.5.0.0
./x86_64-conda-linux-gnu/lib/libgfortran.so
./lib/gcc/x86_64-conda-linux-gnu/14.2.0/libhwasan.so
```

Those unresolved symlinks are expected (unchanged from before cleaning).

Guidance that this is safe:
- https://stackoverflow.com/questions/56266229/is-it-safe-to-manually-delete-all-files-in-pkgs-folder-in-anaconda-python
- https://web.archive.org/web/20240404213219/https://groups.google.com/a/anaconda.com/g/anaconda/c/xV1BiGPmgao

### Don't include rust

I added `rust` in #207 so that we could build `codecov-cli` from source for Python 3.13. This greatly increased our image size because `rust` ships a TON of documentation, including 750 MB of HTML.

```
/opt/conda/share/doc/rust# du -sh *
...
691M    html
56M     json
```

There are also a few other dependencies that we can eliminate by uninstalling rust. Here is the before/after of uninstalling `rust`:

```
/opt/conda# du -sh .
2.6G    .
/opt/conda# mamba uninstall rust
...
/opt/conda# du -sh .
639M    .
```

### DIFF
--- a/citestwheel.Dockerfile
+++ b/citestwheel.Dockerfile
@@ -153,7 +153,7 @@ pyenv rehash
 EOF
 
 # Install latest gha-tools
-RUN -q wget https://github.com/rapidsai/gha-tools/releases/latest/download/tools.tar.gz -O - \
+RUN wget -q https://github.com/rapidsai/gha-tools/releases/latest/download/tools.tar.gz -O - \
   | tar -xz -C /usr/local/bin
 
 # git safe directory
--- a/citestwheel.Dockerfile
+++ b/citestwheel.Dockerfile
@@ -101,7 +101,7 @@ case "${LINUX_VER}" in
     update-ca-trust extract
     dnf clean all
     pushd tmp
-    wget https://www.openssl.org/source/openssl-1.1.1k.tar.gz
+    wget -q https://www.openssl.org/source/openssl-1.1.1k.tar.gz
         tar -xzvf openssl-1.1.1k.tar.gz
     cd openssl-1.1.1k
     ./config --prefix=/usr --openssldir=/etc/ssl --libdir=lib no-shared zlib-dynamic
@@ -120,7 +120,7 @@ EOF
 ARG GH_CLI_VER=notset
 RUN <<EOF
 set -e
-wget https://github.com/cli/cli/releases/download/v${GH_CLI_VER}/gh_${GH_CLI_VER}_linux_${CPU_ARCH}.tar.gz
+wget -q https://github.com/cli/cli/releases/download/v${GH_CLI_VER}/gh_${GH_CLI_VER}_linux_${CPU_ARCH}.tar.gz
 tar -xf gh_*.tar.gz
 mv gh_*/bin/gh /usr/local/bin
 rm -rf gh_*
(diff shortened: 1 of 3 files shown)


bdice:
Before this PR:
```
$ docker run -it rapidsai/ci-conda:latest bash -c 'du -sh /opt/conda'
2.7G    /opt/conda
```
After this PR (local run):
```
$ docker run -it a9c322 bash -c 'du -sh /opt/conda'
641M    /opt/conda
```

citestwheel.Dockerfile
```diff
@@ -153,7 +153,7 @@ pyenv rehash
 EOF
 
 # Install latest gha-tools
-RUN wget https://github.com/rapidsai/gha-tools/releases/latest/download/tools.tar.gz -O - \
+RUN -q wget https://github.com/rapidsai/gha-tools/releases/latest/download/tools.tar.gz -O - \
```
jameslamb, line 156:
```suggestion
RUN wget -q https://github.com/rapidsai/gha-tools/releases/latest/download/tools.tar.gz -O - \
```

Looks like this was a mistake.

msarahan:
Awesome work!  The rules with conda are:

* If any prefix replacement happens, then you get an altered copy of the file in the package cache
* There are no hard links across volumes. Any cross-volume boundary implies a copy.
* Otherwise, everything is hard-linked and does not take any extra space, even though it appears in two places

There are no symlinks back to the package cache. There are only hard links. https://www.redhat.com/en/blog/linking-linux-explained

msarahan (approved):

bdice:
@jameslamb @vyasr Do you have an idea why the `pyenv update` is failing? I haven't used `pyenv` before but I think both of you have (apologies if not).

vyasr:
I'm not sure. That looks like a pyenv bug to me. https://github.com/pyenv/pyenv-update/issues/22 seems like the most relevant issue, but that's ages old and I don't know if the update command's logic has been upstreamed into the main pyenv repo at this point. That issue could be a good starting point for debugging though.

jameslamb:
I'm making progress on it. Documented in #257, testing in #258

jameslamb:
I was able to reproduce the failures in #257 and just merged #258 which appears to fix them. Just merged in latest `main` here, hopefully everything will pass an we can merge this.

gforsyth (approved):

Issue #174, opened 2024-07-09
Use Conda to install CUDA Toolkit (for 12.x)


jakirkham:
Now that the CUDA Toolkit is fully supplied via Conda packages starting with CUDA 12.0, there is no need to depend on the `nvidia/cuda` images (except for CUDA 11.x)

Given this, it might make sense to change the `FROM` logic here to simply use the Linux distro image we would like for CUDA 12.x

https://github.com/rapidsai/miniforge-cuda/blob/0e20886991c4fcb178d73bd0f1cf3224d1b0e0a8/Dockerfile#L3

The CUDA 11.x case would still need the `nvidia/cuda` images

vyasr:
The other benefit is ensuring that there is no risk of accidentally finding the wrong copy of the library, which can still happen if a build tool messes up. I agree that we won't reduce size much further with this change. The speed argument makes sense, although note that we'll still be blocked on RAPIDS overall since we need the images for building our wheel images so it just means there are more things that we can get done sooner in our timeline.

bdice:
> we'll still be blocked on RAPIDS overall since we need the images for building our wheel images

The logic needed here to recreate the `nvidia/cuda` base could be extended a tiny bit to recreate larger images suitable for building wheels. We should start with ci-conda but eliminating the `nvidia/cuda` dependency entirely should be totally viable.

vyasr:
You're basically suggesting that we install the CTK manually into lower-level base images rather than use the `nvidia/cuda` images, and thus avoid `nvidia/cuda` altogether for both wheels and conda, is that right? I would be fine with that.

bdice:
Yes, that's what I am proposing.

Here is a prototype for `ci-conda`. I would welcome thoughts from Docker experts: https://github.com/rapidsai/ci-imgs/pull/266

bdice:
Upon prototyping this, I think the major benefit is that we aren't tied to `nvidia/cuda` base images and thus never have to worry about waiting on images or having unsupported OS versions (we are only tied to the CUDA support matrix and not the `nvidia/cuda` support matrix). The major cost is that we have to duplicate a lot of logic that is present in the `nvidia/cuda` Dockerfiles, and we must then bear the maintenance burden of certain changes, such as updates to how CUDA packages are distributed.

bdice:
I decided not to move forward with this due to reasons mentioned here: https://github.com/rapidsai/ci-imgs/pull/266#issuecomment-2855535057

PR #266, opened 2025-04-28
Drop nvidia/cuda image dependency from ci-conda

bdice:
xref: #174

This needs to be throughly tested with downstream repos before merging.

### DIFF
--- a/ci-conda.Dockerfile
+++ b/ci-conda.Dockerfile
@@ -15,11 +15,15 @@ ARG LINUX_VER
 ARG PYTHON_VER
 ARG CUDA_VER
 ARG CPU_ARCH
+ARG NVARCH
 ARG DEBIAN_FRONTEND=noninteractive
 ENV PATH=/opt/conda/bin:$PATH
 ENV PYTHON_VERSION=${PYTHON_VER}
 
 SHELL ["/bin/bash", "-euo", "pipefail", "-c"]
+
+ENV NVARCH=${NVARCH}
+ENV CUDA_VERSION=${CUDA_VER}
 
 # Install CUDA based on the nvidia/cuda templates
 RUN <<EOF
@@ -28,7 +32,7 @@ case "${LINUX_VER}" in
     # Extract Ubuntu version number using bash variable expansion
     UBUNTU_VERSION=${LINUX_VER#ubuntu}
     UBUNTU_VERSION=${UBUNTU_VERSION//./}
-    CUDA_VER_DASHED=$(echo ${CUDA_VER} | tr '.' '-' | cut -d'-' -f1-2)
+    PKG_CUDA_VER="$(echo ${CUDA_VER} | cut -d '.' -f1,2 | tr '.' '-')"
 
     # Install CUDA for Ubuntu
     apt-get update && apt-get install -y --no-install-recommends gnupg2 curl ca-certificates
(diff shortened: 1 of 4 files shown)


ci-conda.Dockerfile
```diff
@@ -1,21 +1,95 @@
...
+    UBUNTU_VERSION=${UBUNTU_VERSION//./}
+    PKG_CUDA_VER="$(echo ${CUDA_VER} | cut -d '.' -f1,2 | tr '.' '-')"
+
+    # Install CUDA for Ubuntu
+    apt-get update && apt-get install -y --no-install-recommends gnupg2 curl ca-certificates
+    curl -fsSL "https://developer.download.nvidia.com/compute/cuda/repos/ubuntu${UBUNTU_VERSION}/${NVARCH}/3bf863cc.pub" | apt-key add -
+    echo "deb https://developer.download.nvidia.com/compute/cuda/repos/ubuntu${UBUNTU_VERSION}/${NVARCH} /" > /etc/apt/sources.list.d/cuda.list
+    apt-get purge --autoremove -y curl
+    apt-get update && apt-get install -y --no-install-recommends \
+      cuda-cudart-${PKG_CUDA_VER} \
```
bdice, line 43: I'm not sure if we even need `cuda-cudart` here. Maybe just `cuda-compat`?

bdice:
It seems like all the `ci-wheel` and `citestwheel` images are failing to pull `libcudnn8-devel`? I am not sure why, since I don't think I modified anything touching that in this PR.

ci/compute-build-args.sh
```diff
@@ -17,15 +16,49 @@ if [[
   MANYLINUX_VER="manylinux_2_31"
 fi
 
-# translate ARCH to conda-equivalent string values
-CONDA_ARCH=$(echo "$ARCH" | sed 's#amd64#linux64#' | sed 's#arm64#aarch64#')
+
+# Set BASE_IMAGE based on LINUX_VER
+if [[ "${IMAGE_REPO}" == "ci-conda" ]]; then
```
bdice, line 21:
```suggestion
if [[ "${IMAGE_REPO}" == "miniforge-cuda" || "${IMAGE_REPO}" == "ci-conda" ]]; then
```

bdice:
I'm working on an extension of this for wheels, which should be ready soon, but I think we should try to merge this one first if possible.

ci-conda.Dockerfile
```diff
@@ -1,21 +1,95 @@
...
+    ;;
+esac
+EOF
+
+# Set environment variables to match nvidia/cuda image
+ENV PATH=/usr/local/cuda/bin:${PATH}
+ENV LD_LIBRARY_PATH=/usr/local/cuda/lib64
+
+# Required for nvidia-docker v1
+RUN echo "/usr/local/cuda/lib64" >> /etc/ld.so.conf.d/nvidia.conf
```
AyodeAwe, line 87: This command can probably go in the RUN block above it since's there's a logical relation

ci-conda.Dockerfile
```diff
@@ -1,21 +1,95 @@
...
+ENV CUDA_VERSION=${CUDA_VER}
+
+# Install CUDA based on the nvidia/cuda templates
+RUN <<EOF
+case "${LINUX_VER}" in
+  "ubuntu"*)
+    # Extract Ubuntu version number using bash variable expansion
+    UBUNTU_VERSION=${LINUX_VER#ubuntu}
+    UBUNTU_VERSION=${UBUNTU_VERSION//./}
+    PKG_CUDA_VER="$(echo ${CUDA_VER} | cut -d '.' -f1,2 | tr '.' '-')"
```
AyodeAwe, line 35: Minor, but this is declared twice with the same logic, can we declare it once in the RUN block?

jameslamb:
> This needs to be throughly tested with downstream repos before merging.

Noting for the GitHub history something I mentioned offline... some images built here are used as bases for the user-facing RAPIDS container images built in https://github.com/rapidsai/docker.

Specifically, `rapidsai/base` uses `rapidsai/miniforge-cuda` as its base image:

https://github.com/rapidsai/docker/blob/395af8e7b7ec55fdc72472d69041e963f3525489/Dockerfile#L43

`rapidsai/miniforge-cuda` comes from this repo:

https://github.com/rapidsai/ci-imgs/blob/7ff9ab59881f15ace5116916334a9bda65d5926f/ci-conda.Dockerfile#L9-L102

bdice:
Due to the concerns about user-facing images above and the challenges in accurately recreating `NVIDIA_REQUIRE_CUDA` values from `nvidia/cuda` images, we are going to drop this. We will instead wait for `nvidia/cuda` images to release for new CUDA versions and use those, as we have done in the past.

PR #270, opened 2025-05-12
refactor: set Python 3.13 as latest

gforsyth:
xref rapidsai/build-planning#120


bdice (approved):
Yay!

gforsyth:
Oh, I guess this might break `cuopt`?

gforsyth:
Wait, doesn't matter.  They can always point to an explicit 3.12 image.  Merging!

PR #272, opened 2025-05-27
Add pre-commit config

bdice:
This adds a basic `.pre-commit-config.yaml`. We can add more hooks over time as needed.


jameslamb (approved):
great, thank you!!

PR #273, opened 2025-05-29
refactor: remove cuda11 channels and enable strict channel priority

gforsyth:
The `dask/label/dev` channel isn't used anymore, and only a few CUDA 11 packages required the `nvidia` channel, so removing both of those.  And then we're enabling strict priority for the wheels.

### DIFF
--- a/context/condarc.tmpl
+++ b/context/condarc.tmpl
@@ -2,9 +2,7 @@ auto_update_conda: False
 channels:
   - rapidsai
   - rapidsai-nightly
-  - dask/label/dev
   - conda-forge
-  - nvidia
 conda-build:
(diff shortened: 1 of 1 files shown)


bdice (commented):
@gforsyth Are we sure we don’t need these for supporting CUDA 11 in the 25.06 release? We have to wait to drop CUDA 11 from this repo until after 25.06 has shipped.

gforsyth:
We can certainly revert it.  I gather code freeze on a release branch doesn't include pinning the CI images?

bdice:
Correct. We don’t have a branching system for this repo. We have to wait until the previous release is done to make breaking changes in ci-imgs.

PR #278, opened 2025-06-04
Downgrade cuda-compat-12-8 due to upstream bug

bdice:
@gmarkall reported an issue with `cuda-compat-12-8`. This can be worked around temporarily by installing `cuda-compat-12-8=570.148.08-0ubuntu1`.

cc: @jakirkham @kkraus14

### DIFF
--- a/citestwheel.Dockerfile
+++ b/citestwheel.Dockerfile
@@ -78,7 +78,7 @@ case "${LINUX_VER}" in
     update-ca-certificates
 
     # Downgrade cuda-compat on CUDA 12.8 due to an upstream bug
-    if [[ "${RAPIDS_CUDA_VERSION}" == "12.8"* ]]; then
+    if [[ "${CUDA_VER}" == "12.8"* ]]; then
       apt-get install -y cuda-compat-12-8=570.148.08-0ubuntu1
     fi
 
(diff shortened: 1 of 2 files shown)


citestwheel.Dockerfile
```diff
@@ -76,6 +76,12 @@ case "${LINUX_VER}" in
       xz-utils \
       zlib1g-dev
     update-ca-certificates
+
+    # Downgrade cuda-compat on CUDA 12.8 due to an upstream bug
+    if [[ "${RAPIDS_CUDA_VERSION}" == "12.8"* ]]; then
```
kkraus14, line 81: Would probably be annoying, but might be worthwhile to check for the explicit version of the broken cuda-compat package so that when a fix is released we can float forward onto it?

kkraus14 (approved):

citestwheel.Dockerfile
bdice: 570.148.08-0ubuntu1 is the latest working version. The latest two versions are broken.
```
# apt-cache show cuda-compat-12-8 | grep -E "(Version|Installed-Size)"
Version: 575.57.08-0ubuntu1
Installed-Size: 30
Version: 570.148.08-1ubuntu1
Installed-Size: 30
Version: 570.148.08-0ubuntu1
Installed-Size: 197134
Version: 570.133.20-0ubuntu1
Installed-Size: 197130
Version: 570.124.06-0ubuntu1
Installed-Size: 197138
Version: 570.86.15-0ubuntu1
Installed-Size: 196950
Version: 570.86.10-0ubuntu1
Installed-Size: 196950
```

PR #283, opened 2025-06-10
Drop CUDA 11

bdice:
This PR drops CI images for CUDA 11. RAPIDS no longer supports CUDA 11 in the 25.08 release. xref: https://github.com/rapidsai/build-planning/issues/184


bdice:
@kkraus14 @leofang Please let us know if dropping CUDA 11 CI images will affect numba-cuda or cuda-python teams. Happy to discuss details offline.

leofang:
The cuda-python repo does not use any docker image.

The numba-cuda repo does still use the ci-conda 11.8 image. I'll defer to Keith and @gmarkall. 
https://github.com/NVIDIA/numba-cuda/blob/5625de86e50aa918301e53f3e22ad7723f444e17/.github/workflows/conda-python-tests.yaml#L84

gmarkall:
I think it sounds fine given that as I understand it, CUDA 11.8 images will continue to exist, just not get updates / fixes. If we do discover any issue with it during the time in which we continue to support CUDA 11, I think we will have to work around it ourselves - it sounds relatively unlikely to be an issue, so I think we are fine from the Numba-CUDA perspective.

PR #285, opened 2025-06-16
refactor: use `retry` scripts to retry all network cmds

gforsyth:
It's imperfect, but this might help with the `ci-imgs` CI completing more regularly.

Each of the current Dockerfiles installs `gha-tools` at some point -- this PR moves that step to the top of the file, then uses `rapids-retry`, `rapids-pip-retry`, etc to retry any of the docker steps that hit the network (the steps most likely to flake out).

The install of `gha-tools` itself is still prone to failing due to network issues, but this should still reduce the number of failures in the build and publish process.


bdice (commented):
@gforsyth Nice. We are thinking alike. Can we adopt the approach in #288 of installing gha-tools first? I don’t see a reason for us to bundle rapids-retry and then reinstall it at the end. I can add retries for the curl and yum commands as you did here.

gforsyth:
Closing in favor of #288

PR #288, opened 2025-06-18
Use gha-tools retry scripts

bdice:
This PR installs `gha-tools` first, rather than last, and ensures that conda/pip commands use the retrying equivalent to minimize risk of failures.

A [recent failure](https://github.com/rapidsai/ci-imgs/actions/runs/15721271197/job/44302836709#step:9:2040) looked like:

```
 15.66 CondaHTTPError: HTTP 502 BAD GATEWAY for url <https://conda.anaconda.org/conda-forge/linux-aarch64/yaml-cpp-0.8.0-h5ad3122_0.conda>
 15.66 Elapsed: 00:03.035743
 15.66 CF-RAY: 9516dd6ecb952a9c-CMH
 15.66
 15.66 An HTTP error occurred when trying to retrieve this URL.
 15.66 HTTP errors are often intermittent, and a simple retry will get you on your way.
```

### DIFF
--- a/ci-conda.Dockerfile
+++ b/ci-conda.Dockerfile
@@ -19,6 +19,18 @@ ENV PYTHON_VERSION=${PYTHON_VER}
 
 SHELL ["/bin/bash", "-euo", "pipefail", "-c"]
 
+# Set apt policy configurations
+RUN <<EOF
+case "${LINUX_VER}" in
+  "ubuntu"*)
+    echo 'APT::Update::Error-Mode "any";' > /etc/apt/apt.conf.d/warnings-as-errors
+    echo 'APT::Acquire::Retries "10";' > /etc/apt/apt.conf.d/retries
+    echo 'APT::Acquire::https::Timeout "240";' > /etc/apt/apt.conf.d/https-timeout
+    echo 'APT::Acquire::http::Timeout "240";' > /etc/apt/apt.conf.d/http-timeout
+    ;;
+esac
+EOF
+
 # Install latest gha-tools
 RUN <<EOF
 case "${LINUX_VER}" in
@@ -109,9 +121,9 @@ case "${LINUX_VER}" in
         tzdata_pkgs=(tzdata)
     fi
 
-    rapids-retry apt-get update
-    rapids-retry apt-get upgrade -y
-    rapids-retry apt-get install -y --no-install-recommends \
+    apt-get update
+    apt-get upgrade -y
+    apt-get install -y --no-install-recommends \
       "${tzdata_pkgs[@]}"
 
     # Downgrade cuda-compat on CUDA 12.8 due to an upstream bug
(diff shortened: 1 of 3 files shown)


gforsyth:
I did something similar in https://github.com/rapidsai/ci-imgs/pull/285 if you want to take a look

ci-conda.Dockerfile
```diff
@@ -19,6 +19,10 @@ ENV PYTHON_VERSION=${PYTHON_VER}
 
 SHELL ["/bin/bash", "-euo", "pipefail", "-c"]
 
+# Install latest gha-tools
+RUN wget -q https://github.com/rapidsai/gha-tools/releases/latest/download/tools.tar.gz -O - \
+  | tar -xz -C /usr/local/bin
+
```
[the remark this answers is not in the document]
bdice: wget is installed kind of late for some of the images. I added a block to install wget, download gha-tools, and purge wget.

ci-conda.Dockerfile
```diff
@@ -87,9 +109,9 @@ case "${LINUX_VER}" in
         tzdata_pkgs=(tzdata)
     fi
 
-    apt-get update
-    apt-get upgrade -y
-    apt-get install -y --no-install-recommends \
+    rapids-retry apt-get update
+    rapids-retry apt-get upgrade -y
+    rapids-retry apt-get install -y --no-install-recommends \
```
bdice, line 114: I added `rapids-retry` to all `apt-get` commands that happen _before_ we set up the apt retry policy. For `apt-get` commands after the apt retry policy is set, I didn't bother.

Should we instead move the apt retry policy to the top of the Dockerfile, so that we never have to prefix `rapids-retry apt-get` and rely on apt's own retry system?

ci-conda.Dockerfile
```diff
@@ -101,8 +123,8 @@ case "${LINUX_VER}" in
     rm -rf "/var/lib/apt/lists/*"
     ;;
   "rockylinux"*)
-    yum update -y
-    yum clean all
+    dnf update -y
```
bdice, line 126: Aligning these on `dnf`, we have had a mixture of `yum` and `dnf` but it seems that `dnf` is the preferred / modern command to use.

ci-conda.Dockerfile
gforsyth: And if we set the `apt` retry settings first, we'll also reduce the chance of that initial install of `wget` failing

gforsyth: case in point: https://github.com/rapidsai/ci-imgs/actions/runs/15734452777/job/44344936107?pr=288

ci-conda.Dockerfile
```diff
@@ -19,6 +19,40 @@ ENV PYTHON_VERSION=${PYTHON_VER}
 
 SHELL ["/bin/bash", "-euo", "pipefail", "-c"]
 
+# Set apt policy configurations
```
gforsyth, line 22:
```suggestion
# Set apt policy configurations
# We bump up the number of retries and the timeouts for `apt`
# Note that `dnf` defaults to 10 retries, so no additional configuration is required here
```

I thought some extra comments here would help

citestwheel.Dockerfile
```diff
@@ -29,14 +29,44 @@ ENV PATH="/pyenv/bin:/pyenv/shims:$PATH"
 
 SHELL ["/bin/bash", "-euo", "pipefail", "-c"]
 
+# Set apt policy configurations
```
gforsyth, line 32:
```suggestion
# Set apt policy configurations
# We bump up the number of retries and the timeouts for `apt`
# Note that `dnf` defaults to 10 retries, so no additional configuration is required here
```

gforsyth (approved):
This looks good to me -- we could expand the comments at the top to explain why we configure `apt` but not `dnf`, but not blocking.

PR #290, opened 2025-06-18
Update to CUDA 12.9.1.

bdice:
Updates to CUDA 12.9.1.

xref: https://github.com/rapidsai/build-planning/issues/195


gforsyth (approved):

ci-conda.Dockerfile
```diff
@@ -37,7 +37,7 @@ EOF
 RUN <<EOF
 case "${LINUX_VER}" in
   "ubuntu"*)
-    apt-get update && apt-get install -y wget
+    rapids-retry apt-get update -y && apt-get install -y wget
```
bdice, line 40: Package lists are failing to download. I think `APT::Acquire::Retries` only applies to files, not package lists. Retrying more aggressively, in the spirit of #285 / #288.

ci-conda.Dockerfile
```diff
@@ -37,7 +37,8 @@ EOF
 RUN <<EOF
 case "${LINUX_VER}" in
   "ubuntu"*)
-    apt-get update && apt-get install -y wget
+    i=0; until apt-get update -y; do ((++i >= 5)) && break; sleep 10; done
+    apt-get install -y wget
```
jameslamb, line 41:
```suggestion
    apt-get install -y --no-install-recommends wget
```

Could we please add a `--no-install-recommends` here, to avoid pulling in unnecessary packages?

ci-wheel.Dockerfile
```diff
@@ -51,7 +51,8 @@ EOF
 RUN <<EOF
 case "${LINUX_VER}" in
   "ubuntu"*)
-    apt-get update && apt-get install -y wget
+    i=0; until apt-get update -y; do ((++i >= 5)) && break; sleep 10; done
+    apt-get install -y wget
```
jameslamb, line 55:
```suggestion
    apt-get install -y --no-install-recommends wget
```

citestwheel.Dockerfile
```diff
@@ -47,7 +47,8 @@ EOF
 RUN <<EOF
 case "${LINUX_VER}" in
   "ubuntu"*)
-    apt-get update && apt-get install -y wget
+    i=0; until apt-get update -y; do ((++i >= 5)) && break; sleep 10; done
+    apt-get install -y wget
```
jameslamb, line 51:
```suggestion
    apt-get install -y --no-install-recommends wget
```

ci-conda.Dockerfile
```diff
@@ -37,7 +37,8 @@ EOF
 RUN <<EOF
 case "${LINUX_VER}" in
   "ubuntu"*)
-    apt-get update && apt-get install -y wget
+    i=0; until apt-get update -y; do ((++i >= 5)) && break; sleep 10; done
```
jameslamb, line 40: Why are some of these using this `until` loop and others using `rapids-retry`?

If it was to customize the number of retries or sleep time, both can be controlled with shell variables.

```shell
RAPIDS_RETRY_MAX=5 \
RAPIDS_RETRY_SLEEP=10 \
  rapids-retry apt-get update -y
```

source: https://github.com/rapidsai/gha-tools/blob/f25369682c8062d497803e72c4dd5815892d1fad/tools/rapids-retry#L49-L51

jameslamb (approved):
Left some small recommendations for your consideration.

ci-conda.Dockerfile
gforsyth: This line is just before we install `wget` which we then use to install `gha-tools`, so for this particular instance we need to loop manually

jakirkham:
James could you please send these in a new PR?

That way we can make sure they get in

jameslamb:
> _James could you please send these in a new PR?_

Bradley did that already: https://github.com/rapidsai/ci-imgs/pull/292

Looks like this PR just wasn't linked from there.

PR #291, opened 2025-06-20
Rehash after installing anaconda

vyasr:
Without this pyenv does not guarantee that shell commands will be available post installation. It looks like we have been getting lucky so far (perhaps some of the retrying changes changed some subtle pip ordering that made this work before) but the rehash is the right solution. All of RAPIDS CI is currently failing to upload wheels without this.


vyasr:
Example failure: https://github.com/rapidsai/rmm/actions/runs/15771637251/job/44457811186

jameslamb (approved):

PR #298, opened 2025-08-05
double-tag images (with and without RAPIDS version)

jameslamb:
Contributes to https://github.com/rapidsai/build-planning/issues/187

#287 added the RAPIDS version to image tags, like this:

```text
# before
rapidsai/citestwheel:cuda12.8.0-ubuntu24.04-py3.10

# after
rapidsai/citestwheel:25.08-cuda12.8.0-ubuntu24.04-py3.10
```

This allows us to merge breaking changes to the images for one RAPIDS release without impacting the ability to build and test older releases, which is great 🤩 

But it also means that other projects depending on these images which are not as tightly coupled to RAPIDS now have to think about what RAPIDS release(s) are active.

This proposes restoring the un-RAPIDS-versioned images via double-tagging. Please see https://github.com/rapidsai/build-planning/issues/187#issuecomment-2981069294 for details.

### How I tested this

Tested the GitHub Actions changes with `actionlint` (which did catch a mistake I'd made!)

```shell
actionlint .github/workflows/*
```

Checked the matrix entries like this:

```shell
# with https://github.com/rapidsai/gha-tools cloned to ~/repos/gha-tools

PATH="${HOME}/repos/gha-tools/tools:${PATH}" \
BUILD_TYPE=branch \
  ./ci/compute-matrix.sh \
| jq .
```

Saw the same number of total images, but a new `IMAGE_NAME_NO_RAPIDS_VERSION` added to each matrix entry (with correct values).

```json
{
  [
    ...
    {
      "CUDA_VER": "12.9.1",
      "PYTHON_VER": "3.13",
      "LINUX_VER": "rockylinux8",
      "IMAGE_REPO": "citestwheel",
      "DOCKERFILE": "citestwheel.Dockerfile",
      "DOCKER_TARGET": "",
      "ARCHES": [
        "amd64",
        "arm64"
      ],
      "IMAGE_NAME": "rapidsai/citestwheel:25.10-cuda12.9.1-rockylinux8-py3.13",
      "IMAGE_NAME_NO_RAPIDS_VERSION": "rapidsai/citestwheel:cuda12.9.1-rockylinux8-py3.13"
    }
  ]
}
```

### DIFF
--- a/.github/workflows/build-and-publish-images.yaml
+++ b/.github/workflows/build-and-publish-images.yaml
@@ -80,4 +80,5 @@ jobs:
           IMAGE_REPO: ${{ matrix.IMAGE_REPO }}
           BUILD_TYPE: ${{ inputs.build_type }}
           IMAGE_NAME: ${{ matrix.IMAGE_NAME }}
+          IMAGE_NAME_NO_RAPIDS_VERSION: ${{ matrix.IMAGE_NAME_NO_RAPIDS_VERSION }}
         run: ci/create-multiarch-manifest.sh
--- a/ci/compute-matrix.jq
+++ b/ci/compute-matrix.jq
@@ -23,6 +23,16 @@ def compute_tag_prefix($x):
     $x.IMAGE_REPO + "-" + env.PR_NUM + "-" + env.RAPIDS_VERSION_MAJOR_MINOR + "-"
   end;
 
+def compute_tag_prefix_no_rapids_version($x):
+  if
+    env.BUILD_TYPE == "branch"
+  then
+    ""
+  else
+    $x.IMAGE_REPO + "-" + env.PR_NUM + "-"
+  end;
+
+# Compute image URI in the form '{repo}/{name}:{tag}'
 def compute_image_name($x):
   compute_repo($x) as $repo |
   compute_tag_prefix($x) as $tag_prefix |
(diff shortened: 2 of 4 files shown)


.github/workflows/build-and-publish-images.yaml
```diff
@@ -80,4 +80,5 @@ jobs:
           IMAGE_REPO: ${{ matrix.IMAGE_REPO }}
           BUILD_TYPE: ${{ inputs.build_type }}
           IMAGE_NAME: ${{ matrix.IMAGE_NAME }}
+          IMAGE_NAME_NO_RAPIDS_VERSION: ${{ matrix.IMAGE_NAME_NO_RAPIDS_VERSION }}
```
jameslamb, line 83: This is only needed in the jobs that are pushing new manifests, because I'm not proposing also double-publishing the architecture-specific tags like `rapidsai/ci-conda:[25.08-]cuda12.8.0-ubuntu24.04-py3.10-arm64`.

Those tags can technically be pulled directly, but they're really just intended as implementation details of the non-architecture-specific manifests, e.g. `rapidsai/ci-conda:[25.08-]cuda12.8.0-ubuntu24.04-py3.10`

ci/compute-matrix.jq
```diff
@@ -23,6 +23,16 @@ def compute_tag_prefix($x):
...
   end;
 
+def compute_tag_prefix_no_rapids_version($x):
+  if
+    env.BUILD_TYPE == "branch"
+  then
+    ""
+  else
+    $x.IMAGE_REPO + "-" + env.PR_NUM + "-"
+  end;
```
jameslamb, line 33: Duplicating these functions with `_no_rapids_version` forms is not the most elegant way to do this, but it was easy to write (for me, someone not that familiar with `jq`) and I think it's easy to read. Happy to try a different way if reviewers are uncomfortable with this logic being duplicated.

jameslamb:
I think this is working! For example, look at https://hub.docker.com/r/rapidsai/staging/tags?name=citestwheel-298

Screenshot 2025-08-06 at 9 14 03 AM

^ 2 tags, one with RAPIDS version and 1 without, pointing to identical images

PR #301, opened 2025-08-21
Install yq and awscli from binaries instead of Docker

jayavenkatesh19:
Towards https://github.com/rapidsai/build-infra/issues/288

Instead of installing AWS CLI and yq from Docker images as base, these changes download the official binaries for both these packages and install them, reducing the number of base images being used in the Docker build.

### DIFF
--- a/ci-conda.Dockerfile
+++ b/ci-conda.Dockerfile
@@ -5,8 +5,6 @@
 ARG CUDA_VER=notset
 ARG LINUX_VER=notset
 ARG PYTHON_VER=notset
-ARG YQ_VER=notset
-ARG AWS_CLI_VER=notset
 ARG MINIFORGE_VER=notset
 
 FROM condaforge/miniforge3:${MINIFORGE_VER} AS miniforge-upstream
@@ -143,10 +141,6 @@ case "${LINUX_VER}" in
 esac
 EOF
 
-FROM mikefarah/yq:${YQ_VER} AS yq
-
-FROM amazon/aws-cli:${AWS_CLI_VER} AS aws-cli
-
 FROM miniforge-cuda
 
 ARG TARGETPLATFORM=notset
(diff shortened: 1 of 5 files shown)


ci-conda.Dockerfile
```diff
@@ -265,6 +261,17 @@ rapids-retry wget -q https://github.com/cli/cli/releases/download/v${GH_CLI_VER}
 tar -xf gh_*.tar.gz
 mv gh_*/bin/gh /usr/local/bin
 rm -rf gh_*
+
+rapids-retry wget -q https://github.com/mikefarah/yq/releases/download/v${YQ_VER}/yq_linux_${CPU_ARCH} -O /tmp/yq
+mv /tmp/yq /usr/bin/yq
+chmod +x /usr/bin/yq
+rm -f /tmp/yq
+
+rapids-retry curl -o /tmp/awscliv2.zip \
```
jameslamb, line 270:
```suggestion
# ref: https://docs.aws.amazon.com/cli/latest/userguide/getting-started-install.html#getting-started-install-instructions
rapids-retry curl -o /tmp/awscliv2.zip \
```

Let's link to the official docs this came from, to help people reading it in the future to understand where these commands came from. Especially since this is doing a bit more than just downloading a binary.

jameslamb (changes requested):
Thanks!

I'm generally supportive of this. My understanding from @jarmak-nv is that fewer base images (even if they're just used for selective copying of files like this) will make it easier to pass security and compliance scans.

This is not quite ready to merge yet though. In this project we use renovate to get auto-update PRs like #293, and that configuration should be changed now that we'll not pulling docker images.

https://github.com/rapidsai/ci-imgs/blob/d384a1f00e6c58b4b675537eb8177eff565547fa/versions.yaml#L12-L15

https://github.com/rapidsai/ci-imgs/blob/d384a1f00e6c58b4b675537eb8177eff565547fa/renovate.json#L13-L15

Please make sure that's updated appropriately. I don't know the exact syntax offhand, refer to https://docs.renovatebot.com/configuration-options/

jayavenkatesh19:
@jameslamb Made the changes to renovate's configuration. Thanks for highlighting this!

jameslamb:
(and thanks @gforsyth for restarting those failing builds)

PR #304, opened 2025-09-10
Use CUDA 13.0.1

bdice:
This updates to CUDA 13.0 update 1. Images for `nvidia/cuda` are now available.

Conda workflows are already using 13.0.1 (by pinning to `cuda-version=13.0`, which doesn't pin patch releases) and CI hasn't shown any problems, so we have reason to believe this should be non-breaking.


jameslamb (approved):

vyasr:
Just to be safe, let's wait until Monday. In case something does break I'd like it to not be over a long weekend.

PR #306, opened 2025-09-30
Drop CUDA 12.0 from CI images

bdice:
This drops CUDA 12.0 from the CI images.

Depends on https://github.com/rapidsai/shared-workflows/pull/431.

xref: https://github.com/rapidsai/build-planning/issues/223

### DIFF
--- a/matrix.yaml
+++ b/matrix.yaml
@@ -2,7 +2,6 @@
 # All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
 CUDA_VER:
-  - "12.0.1"
   - "12.2.2"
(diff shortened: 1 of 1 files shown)


bdice:
The shared-workflows PR is merged, and I see nothing using these images, so this should be ready to merge.

https://github.com/search?q=org%3Arapidsai+%2Frapidsai%5C%2Fci-.*cuda12%5C.0%2F&type=code
https://github.com/search?q=org%3ANVIDIA+%2Frapidsai%5C%2Fci-.*cuda12%5C.0%2F&type=code

PR #310, opened 2025-10-21
fix(rockylinux): run `dnf makecache` so indexes aren't stale

gforsyth:
There are several 404s in the build jobs here: https://github.com/rapidsai/ci-imgs/actions/runs/18690449618

Some of them resolve on a restart, but the ones that fail hit a 404 for every single mirror, which seems weird.  Adding in `dnf makecache` to try to keep the indexes fresh.


gforsyth:
I don't think this made any difference.

PR #322, opened 2025-11-13
Update to 26.02

AyodeAwe:
This PR updates the repository to version 26.02.

This is part of the 26.02 release burndown process.


bdice (approved):
We need to cut `release/25.12` from the commit prior to this.

PR #324, opened 2025-11-13
fix(deps): revert to `miniforge3:25.3.1-0` and pin `libcurl==8.14.1`

gforsyth:
Ok, after a lot of back and forth and trial and error, we've identified that some combination of `libcurl` and certain versions of either `mamba` or `conda` (contained in the upstream `miniforge3` image) are interacting in such a way as to cause deadlocks in some small percentage of runs.  But it is consistent enough to cause at least a few deadlocks for every PR in this repo.

For now (2025/11/14), I have:
* reverted back to the upstream `miniforge3:25.3.1-0` image
* pinned `libcurl=8.14.1`

The `libcurl` pin is set to the version of `libcurl` that is already installed in the `base` environment of the upstream image, the goal with pinning is to avoid swapping in a new version mid-network-transaction.

Noting here that this is still a _temporary_ fix.  We need to better isolate the root cause (between `libcurl` and `mamba` and `conda` versions) and then report upstream to the appropriate project/projects.

BUT, this does let us update `openssl` and stops the deadlocks, so that's still a plus.

### DIFF
--- a/matrix.yaml
+++ b/matrix.yaml
@@ -20,12 +20,12 @@ CI_IMAGE_CONFIG:
   - IMAGE_REPO: "ci-conda"
     dockerfile: ci-conda.Dockerfile
     docker_target: ""
-  # - IMAGE_REPO: "ci-wheel"
-  #   dockerfile: ci-wheel.Dockerfile
-  #   docker_target: ""
-  # - IMAGE_REPO: "citestwheel"
-  #   dockerfile: citestwheel.Dockerfile
-  #   docker_target: ""
+  - IMAGE_REPO: "ci-wheel"
+    dockerfile: ci-wheel.Dockerfile
+    docker_target: ""
+  - IMAGE_REPO: "citestwheel"
+    dockerfile: citestwheel.Dockerfile
+    docker_target: ""
 exclude:
   # Exclusions from CUDA's OS support matrix
   - LINUX_VER: "ubuntu24.04"


ci-conda.Dockerfile
```diff
@@ -80,7 +80,7 @@ echo 'openssl<3.5.3' >> /opt/conda/conda-meta/pinned
 
 # update everything before other environment changes, to ensure mixing
 # an older conda with newer packages still works well
-rapids-mamba-retry update --all -y -n base
+rapids-mamba-retry update --all -y -n -vvvv base
```
rockhowse: Yup with this you will see where it's hanging... however... the buffer in the web interface won't update until the pipeline completes (is cancled or times out)

I kept thinking it was stuck at the place where you see it stop updating in the web console, but after termination and the buffers are flushed you can see it's actually terminating when conda's setting up to connect to conda-forge

Image

gforsyth:
Wow, ok, so reverting to the upstream `miniforge3:25.3.1-0` image and pinning `openssl` and `libcurl` so those packages don't get upgraded in-place in the `base` environment seem to have resolved all of the hangs.

I'm going to try unpinning `openssl` but leaving the `libcurl` pin to see if that also works.

But for now, the things in the mix in re: the hanging behavior are:

* `openssl` (although I think this is a bit of a red herring and it's more that certain `openssl` versions lead to an upgraded `libcurl`)
* `libcurl`
* whatever changes were made to the upstream miniforge image in the 25.9.1-0 release: https://github.com/conda-forge/miniforge/releases/tag/25.9.1-0

Some possible culprits in the upstream release:
* upgraded `conda` and `mamba`: https://github.com/conda-forge/miniforge/pull/827
* new `mirrored_channel` fallback that points to `prefix.dev` mirrors (although I haven't seen any of our workers trying to ping that): https://github.com/conda-forge/miniforge/pull/817

bdice (approved):

rockhowse (approved):
This set of changes is closer to the real issue. Ideally should allow us to have a stable `25.12` release while we address a way for us to use newer versions. 

I agree with @bdice that we want to report this upstream.

PR #332, opened 2025-12-02
Remove sccache

bdice:
We do not want to use the mozilla/sccache (upstream) version of sccache. We should always be using the rapids/sccache fork which contains several required features. That is installed by https://github.com/rapidsai/gha-tools/blob/main/tools/rapids-configure-sccache, which should be called with `source rapids-configure-sccache` in all CI workflows that perform compilation.


bdice:
We can merge this after the gha-tools publish job https://github.com/rapidsai/ci-imgs/actions/runs/19945582004 finishes.

PR #378, opened 2026-03-12
Update latest.yaml CUDA and Python versions to match matrix.yaml

AyodeAwe:
Updates `latest.yaml` to use the greatest CUDA and Python versions from `matrix.yaml`:

- CUDA: `13.1.0` -> `13.1.1`
- Python: `3.13` -> `3.14`

The `-latest` CI image tag is built from the CUDA/Python combo specified in `latest.yaml`. When `matrix.yaml` was updated with new CUDA and Python versions, `latest.yaml` wasn't updated to match, so the `-latest` tag could no longer be published. This broke shared-workflows `@main` which references the `-latest` images.

### DIFF
--- a/latest.yaml
+++ b/latest.yaml
@@ -2,15 +2,15 @@
 # SPDX-License-Identifier: Apache-2.0
 # Define the values used for the "latest" tag
 ci-conda:
-  CUDA_VER: "13.1.0"
-  PYTHON_VER: "3.13"
+  CUDA_VER: "13.1.1"
+  PYTHON_VER: "3.14"
   LINUX_VER: "ubuntu24.04"
 ci-wheel:
-  CUDA_VER: "13.1.0"
-  PYTHON_VER: "3.13"
+  CUDA_VER: "13.1.1"
+  PYTHON_VER: "3.14"
   # Wheels should always be built with the oldest supported glibc version
   LINUX_VER: "rockylinux8"
 citestwheel:
-  CUDA_VER: "13.1.0"
-  PYTHON_VER: "3.13"
+  CUDA_VER: "13.1.1"
+  PYTHON_VER: "3.14"
   LINUX_VER: "ubuntu24.04"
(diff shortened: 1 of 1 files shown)


jameslamb (commented):
> _When matrix.yaml was updated with new CUDA and Python versions, latest.yaml wasn't updated to match, so the -latest tag could no longer be published._

Sorry, I don't understand this comment. Could you link to where "-latest could no longer be published"?

The build at https://github.com/rapidsai/ci-imgs/actions/runs/23013778650 succeeded (though I do see there aren't `26.06-latest` tags at https://hub.docker.com/r/rapidsai/ci-conda/tags)

It's intended that `matrix.yaml` and `latest.yaml` are updated separately.

It's not safe to update `:26.04-latest` to Python 3.14 (for example) until most RAPIDS projects are supporting Python 3.14.

AyodeAwe:
Thanks James. You're right that Python 3.14 shouldn't be bumped here until that work is complete. I've reverted that -- this now only bumps CUDA `13.1.0` -> `13.1.1`.

The issue is that `latest.yaml` had CUDA `13.1.0` but `matrix.yaml` has `13.1.1`, so no matrix entry matched in `create-multiarch-manifest.sh` and the `:latest` / `:26.06-latest` manifests were never created. That's why `26.06-latest` doesn't exist on DockerHub even though the build succeeded -- Brad confirmed the same in [the Slack thread](https://nvidia.slack.com/archives/C01PF84NXCL/p1773340260090849?thread_ts=1773335609.758809&cid=C01PF84NXCL).

jameslamb:
ohhh I see, we do this job-by-job:

https://github.com/rapidsai/ci-imgs/blob/af10fc4c9a0808f2bb836c3f5b7fbb67ee0e2288/ci/create-multiarch-manifest.sh#L31-L35

oy. Thanks for finding that.

Here's a bug report, something to look into in the future: https://github.com/rapidsai/ci-imgs/issues/379

bdice (approved):

PR #380, opened 2026-03-12
Validate latest.yaml entries exist in build matrix

### LINKED ISSUE #379: failure to publish "latest" images should fail CI
## Description

In a recent release, we encountered a situation where no `:latest` or `:{rapids-version}-latest` images were published because none of the matrix entries matched the `latest` configuration.

In other words, every run of this evaluated to False:

https://github.com/rapidsai/ci-imgs/blob/af10fc4c9a0808f2bb836c3f5b7fbb67ee0e2288/ci/create-multiarch-manifest.sh#L31-L35

This should result in a CI failure here.

## Benefits of this work

* saves debugging time... in 26.06 this issue was only found via other projects' CI failing to find a `:26.06-latest` image at runtime

## Acceptance Criteria

* CI passing here means that new `:latest` and `:{rapids-version}-latest` images were successfully published

## Approach

Look at the linked code and the things calling it. Look for stricter patterns that could result in the behavior we want.

Alternatively, a single job that runs after all the manifest-publishing and tries to pull the latest images could have caught this.

## Notes

N/A

bdice:
## Summary

- Adds a `ci/validate-latest.sh` script and workflow step that checks every `latest.yaml` entry has a matching `(IMAGE_REPO, CUDA_VER, PYTHON_VER, LINUX_VER)` tuple in the computed build matrix, failing CI before any builds start if not

Closes #379


jameslamb (approved):
This is a great way to do this!!!

I love the idea of `compute-matrix` having some validation in it. Bet we could extend that to some things in `shared-workflows` in the future to enforce the things we always want to be true ("at least 1 `arm64` job in PR matrix", "at least 1 oldest-dependencies job", etc.).

bdice:
Working:
{4E3E2C08-E271-47BD-8067-B22305DE3A3A}

Failing as expected (with the `latest.yaml` prior to the fix in #378):
{221C1ECF-F404-4963-8552-79ED83AB74BD}

ci/validate-latest.sh
```diff
@@ -0,0 +1,32 @@
...
+set -euo pipefail
+
+MATRIX="${1:-}"
+if [[ -z "${MATRIX}" ]]; then
+  echo "Usage: validate-latest.sh MATRIX_JSON"
+  exit 1
+fi
+
+errors=0
+for image_repo in $(yq -r 'keys | .[]' latest.yaml); do
```
KyleFromNVIDIA: While it probably doesn't specifically matter here, splitting on whitespace is error-prone. Prefer to use null separators:


```suggestion
yq -r -0 'keys | .[]' latest.yaml | while read -r -d $'\0' image_repo; do
```

KyleFromNVIDIA (approved):
Approved with one small comment

jameslamb:
I think it's ok for this to just go straight to `main` and not `release/26.04`. It's important that we have it for future releases but don't need it for 26.04.

PR #382, opened 2026-03-17
Add CUDA 13.2 CI images

jakirkham:
Now that CUDA 13.2 is available. Add CUDA 13.2 CI images.

<hr>

Refs:

* https://github.com/rapidsai/build-planning/issues/265


matrix.yaml
```diff
@@ -4,7 +4,7 @@ CUDA_VER:
   - "12.2.2"
   - "12.9.1"
   - "13.0.2"
-  - "13.1.1"
+  - "13.2.0"
```
jakirkham, line 8: Do we want to keep CUDA 13.1.1 for now or do we want to drop it and move to CUDA 13.2.0?

jakirkham: Perhaps it is useful to maintain both for a period of time until the update to 13.2 is completed?

OTOH it would substantially increase the size of the CI matrix

bdice: Yes, this is important so that 13.1 continues to get CI image updates until we can go back and remove it later once nothing is using it in our build/test workflows.

latest.yaml
bdice, line 1: This file should only be updated at the conclusion of the migration, once all the build/test workflows are using 13.2.

bdice (dismissed):

matrix.yaml
jakirkham: Thanks for the feedback! So like this

```suggestion
  - "13.0.2"
  - "13.1.1"
  - "13.2.0"
```

latest.yaml
jakirkham: Ok got it. Will revert these changes then

latest.yaml
```diff
@@ -2,15 +2,15 @@
 # SPDX-License-Identifier: Apache-2.0
 # Define the values used for the "latest" tag
 ci-conda:
-  CUDA_VER: "13.1.1"
+  CUDA_VER: "13.2.0"
```
jakirkham:
```suggestion
  CUDA_VER: "13.1.1"
```

latest.yaml
```diff
@@ -2,15 +2,15 @@
...
   PYTHON_VER: "3.13"
   LINUX_VER: "ubuntu24.04"
 ci-wheel:
-  CUDA_VER: "13.1.1"
+  CUDA_VER: "13.2.0"
```
jakirkham:
```suggestion
  CUDA_VER: "13.1.1"
```

latest.yaml
```diff
@@ -2,15 +2,15 @@
...
   PYTHON_VER: "3.13"
   # Wheels should always be built with the oldest supported glibc version
   LINUX_VER: "rockylinux8"
 citestwheel:
-  CUDA_VER: "13.1.1"
+  CUDA_VER: "13.2.0"
```
jakirkham:
```suggestion
  CUDA_VER: "13.1.1"
```

jakirkham (commented):
Following up on Bradley's feedback with these proposed changes

matrix.yaml
jakirkham: Done in commit: 6c0ab368102fbfdd6a669dbc7e4cc4d58831bf67

latest.yaml
jakirkham: Done in commit: d0ffc71f63e54bd31e7a12b53ca2fe5a89bfa701

jameslamb (approved):
Fine with me to start building these so we can test things.

jakirkham:
Thank you both for your feedback! 🙏

Merging so we can get these built

PR #383, opened 2026-03-17
feat: set Python 3.14 as latest version

gforsyth:
Part of rapidsai/build-planning#205

Adding a `DO NOT MERGE` tag here so @jameslamb can sign off and confirm this won't mess up his work fixing the dependencies in `cugraph-gnn`


jameslamb (approved):
Yep this should be totally fine! Thanks for the `@`.

If this breaks the docs build in `cugraph-gnn` or something I'll just pin that job to a CUDA 13 image, no worries.

PR #399, opened 2026-04-15
fix: update `pip.conf` location to work around pip config bug

### LINKED ISSUE #398: Change location of global `pip.conf`
In rapidsai/integration#837 I need to unset the `extra-index-url` in the `citestwheel` image.

There seems to be a bug in `pip` about how to handle setting/unsetting flags in global config files:

This is in `rapidsai/citestwheel:cuda12.9.1-ubuntu24.04-py3.14`
```sh
root@1b149050f29d:/# pip --version
pip 26.0.1 from /pyenv/versions/3.14.4/lib/python3.14/site-packages/pip (python 3.14)

root@1b149050f29d:/# pip config --global debug
env_var:
env:
global:
  /etc/xdg/pip/pip.conf, exists: True
    global.quiet: 0
    global.extra-index-url:
    https://pypi.anaconda.org/rapidsai-wheels-nightly/simple
    global.retries: 10
  /etc/pip.conf, exists: False
site:
  /pyenv/versions/3.14.4/pip.conf, exists: False
user:
  /root/.pip/pip.conf, exists: False
  /root/.config/pip/pip.conf, exists: False

root@1b149050f29d:/# pip config --global unset global.extra-index-url
ERROR: Exception:
Traceback (most recent call last):
  File "/pyenv/versions/3.14.4/lib/python3.14/site-packages/pip/_internal/cli/base_command.py", line 107, in _run_wrapper
    status = _inner_run()
  File "/pyenv/versions/3.14.4/lib/python3.14/site-packages/pip/_internal/cli/base_command.py", line 98, in _inner_run
    return self.run(options, args)
           ~~~~~~~~^^^^^^^^^^^^^^^
  File "/pyenv/versions/3.14.4/lib/python3.14/site-packages/pip/_internal/commands/configuration.py", line 139, in run
    handler_map[action](options, args[1:])
    ~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^
  File "/pyenv/versions/3.14.4/lib/python3.14/site-packages/pip/_internal/commands/configuration.py", line 197, in unset_name
    self.configuration.unset_value(key)
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^
  File "/pyenv/versions/3.14.4/lib/python3.14/site-packages/pip/_internal/configuration.py", line 190, in unset_value
    key not in self._config[self.load_only][fname]
               ~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^
KeyError: '/etc/pip.conf'
```

It sees the config file, it acknowledges that it is a valid global location, and those options are used, but you can't use `pip config unset` to remove the option.

As a workaround, if we use `/etc/pip.conf` as our default location:

```sh
root@1b149050f29d:/# mv /etc/xdg/pip/pip.conf /etc/pip.conf

root@1b149050f29d:/# pip config --global debug
env_var:
env:
global:
  /etc/xdg/pip/pip.conf, exists: False
  /etc/pip.conf, exists: True
    global.quiet: 0
    global.extra-index-url:
    https://pypi.anaconda.org/rapidsai-wheels-nightly/simple
    global.retries: 10
site:
  /pyenv/versions/3.14.4/pip.conf, exists: False
user:
  /root/.pip/pip.conf, exists: False
  /root/.config/pip/pip.conf, exists: False

root@1b149050f29d:/# pip config --global unset global.extra-index-url
Writing to /etc/pip.conf

root@1b149050f29d:/# pip config --global list
global.quiet='0'
global.retries='10'
```

I'll check if this is a known bug or not and report upstream, but unless we have a reason to use `/etc/xdg/pip/pip.conf` specifically I think we should move it.

gforsyth:
Resolves #398 -- the two locations _should_ be interchangeable, but they aren't quite, and this location allows us to use `pip config set` and `pip config unset` for plucking out keys we want to selectively remove.

### DIFF
--- a/ci-conda.Dockerfile
+++ b/ci-conda.Dockerfile
@@ -295,6 +295,6 @@ pip cache purge
 EOF
 
 # Add pip.conf
-COPY pip.conf /etc/xdg/pip/pip.conf
+COPY pip.conf /etc/pip.conf
 
 CMD ["/bin/bash"]
--- a/ci-wheel.Dockerfile
+++ b/ci-wheel.Dockerfile
@@ -26,7 +26,7 @@ ENV PATH="${PYENV_ROOT}/bin:${PYENV_ROOT}/shims:$PATH"
 SHELL ["/bin/bash", "-euo", "pipefail", "-c"]
 
 # Add pip.conf
-COPY pip.conf /etc/xdg/pip/pip.conf
+COPY pip.conf /etc/pip.conf
 
 # Install all the tools that are just "download a binary and stick it on PATH".
 #
--- a/citestwheel.Dockerfile
+++ b/citestwheel.Dockerfile
@@ -19,7 +19,7 @@ ENV RAPIDS_CONDA_ARCH="${CONDA_ARCH}"
 SHELL ["/bin/bash", "-euo", "pipefail", "-c"]
 
 # Add pip.conf
-COPY pip.conf /etc/xdg/pip/pip.conf
+COPY pip.conf /etc/pip.conf
 
 # Install all the tools that are just "download a binary and stick it on PATH".
 #


bdice (approved):

jameslamb (approved):
I *think* this should just work, don't see anything in RAPIDS explicitly depending on the `/etc/xdg` path ([GitHub search](https://github.com/search?q=org%3Arapidsai%20%22%2Fetc%2Fxdg%22&type=code))

Let's try it!

jameslamb:
Oh wait sorry... I think we want to update the `ci-conda` one too:

https://github.com/rapidsai/ci-imgs/blob/d38b2a57c33f87e69744cada26cc519b6f1628c4/ci-conda.Dockerfile#L298

That config will still affect `pip` invocations in conda-based jobs, which do happen. One common case I can think of is `checks` jobs, which use conda to set up an environment but then run `pre-commit`... `pre-commit` will invoke `pip` for its environment setup.

gforsyth:
Added the change to `ci-conda.Dockerfile`, too!

PR #400, opened 2026-04-15
refactor(citestwheel): switch base image from devel to base

gforsyth:
Building `citestwheel` using the `base` version of the upstream `cuda` image.

Opening this PR so that the image gets built and pushed to staging - I'll work on testing this out on a few repos before merging.


vyasr:
The main issue with this change is that it precludes us with testing with cupy<14. I'm not sure if that is acceptable yet.

PR #403, opened 2026-05-01
bump to sccache v0.14.0-rapids.12

trxcllnt:
Due to the ongoing Canonical attack, I commented out the `add-apt-repository ppa:git-core/ppa` line in `citestwheel.Dockerfile`. The comment says it's so we install git > v2.17, but Ubuntu 22.04 seems to be our min version, and that has git v2.34.1.

### DIFF
--- a/citestwheel.Dockerfile
+++ b/citestwheel.Dockerfile
@@ -56,9 +56,9 @@ set -e
 case "${LINUX_VER}" in
   "ubuntu"*)
     rapids-retry apt-get update -y
-    apt-get install -y software-properties-common
-    # update git > 2.17
-    add-apt-repository ppa:git-core/ppa -y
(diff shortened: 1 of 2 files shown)


citestwheel.Dockerfile
```diff
@@ -56,9 +56,9 @@ set -e
 case "${LINUX_VER}" in
   "ubuntu"*)
     rapids-retry apt-get update -y
-    apt-get install -y software-properties-common
-    # update git > 2.17
-    add-apt-repository ppa:git-core/ppa -y
+    # apt-get install -y software-properties-common
```
bdice: I think we can delete this block instead of commenting it, I believe it is a relic of Ubuntu 20.04 support.

bdice (approved):

PR #404, opened 2026-05-05
fix(ci): resolve all zizmor findings and add zizmor pre-commit checks

gforsyth:
Similar to upstream changes in `shared-workflows`, this PR cleans up and annotates all of the workflows and adds the `zizmor` linter to make sure changes are checked.

Part of https://github.com/rapidsai/build-planning/issues/275

### DIFF
--- a/.github/workflows/build-image.yaml
+++ b/.github/workflows/build-image.yaml
@@ -1,4 +1,4 @@
-# SPDX-FileCopyrightText: Copyright (c) 2023-2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-FileCopyrightText: Copyright (c) 2023-2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
 name: Build and push image variant
 
--- a/.github/workflows/build-image.yaml
+++ b/.github/workflows/build-image.yaml
@@ -45,7 +45,7 @@ jobs:
     runs-on: linux-${{ matrix.ARCH }}-cpu4
     steps:
       - name: Checkout
-        uses: actions/checkout@v6
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
         with:
           persist-credentials: false
       - name: Download gha-tools with git clone
@@ -53,7 +53,7 @@ jobs:
           git clone https://github.com/rapidsai/gha-tools.git -b main /tmp/gha-tools
           echo "/tmp/gha-tools/tools" >> "${GITHUB_PATH}"
       - name: Login to DockerHub
-        uses: docker/login-action@v4
+        uses: docker/login-action@4907a6ddec9925e35a0a9e82d7399ccc52663121 # v4.1.0
         with:
           username: ${{ secrets.GPUCIBOT_DOCKERHUB_USER }}
           password: ${{ secrets.GPUCIBOT_DOCKERHUB_TOKEN }}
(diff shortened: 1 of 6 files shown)


.github/workflows/build-image.yaml
```diff
@@ -45,21 +45,23 @@ jobs:
     runs-on: linux-${{ matrix.ARCH }}-cpu4
     steps:
       - name: Checkout
-        uses: actions/checkout@v6
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
```
jameslamb:
```suggestion
        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
```

This is `v6.0.2` (https://github.com/actions/checkout/releases/tag/v6.0.2.

I *think* specifying that full tag might be useful for renovate's update logic? I'm not sure, but recommend erring on the side of pinning to a specific semver where it's available.

If nothing else, I think using the full version should mean the diff for the next automatic update will change the comment too, so we'll be able to tell the difference visually between `6.0.2 -> 6.0.3` and `6.0.2 -> 6.1.0`.

Would you consider that for all of these?

.github/workflows/pr.yaml
```diff
@@ -20,8 +24,12 @@ jobs:
...
+    permissions:
+      contents: read
+      id-token: write
+      pull-requests: read
   build-images:
     uses: ./.github/workflows/build-and-publish-images.yaml
     with:
       build_type: pull-request
-    secrets: inherit
+    secrets: inherit # zizmor: ignore[secrets-inherit]
```
jameslamb: Instead of `secrets: inherit`, would you consider explicitly defining which secrets need to get passed through in `workflow_call` calls?

Here's an example of that: https://github.com/rapidsai/shared-workflows/pull/489 (though here we don't need the name and value separated, can just pass through the value directly).

That's tighter than `secrets: inherit` because it means that new secrets that become available to the repo aren't immediately accessible in the calling workflow.

I also think it has the nice side benefit of making the configuration flow a bit more explicit.

jameslamb (approved):
Left some suggestions for your consideration, but I'm also comfortable with this being merged as-is if you think they're not worth implementing.

.github/workflows/build-image.yaml
gforsyth: Yeah, the tags vs. releases are weird -- I think our mutable refs are pointing at tags (one more reason not to use them) -- so it is `6.0.2` but it is also `v6` (https://github.com/actions/checkout/releases/tag/v6) -- and I think they change what `v6` refers to for point releases. I can update that.

.github/workflows/pr.yaml
gforsyth: I agree it's better to explicilty define the secrets -- and because it's currently all implicit it's hard to trace out which secrets are needed.

I might defer this until we can create a general grammar for which shared workflows make use of which secrets and then take a pass to tighten up the secret passing.

.github/workflows/build-image.yaml
jameslamb: I agree with your read, that's exactly right.

`v6` gets updated every time a new `v6.*` is cut.

.github/workflows/pr.yaml
gforsyth: I have more thoughts on this after some spelunking but I think it should get written up somewhere else.

I'll take a pass on this PR, though, and tighten up the secrets usage where possible (it's largely not possible at the moment)

gforsyth: xref rapidsai/build-infra#358

jameslamb: Latest changes + that issue look good, thank you for that!

jakirkham:
Pulled the latest changes into this PR

This also reruns CI. Hoping this clears out [the previous `apt-get` repo availability issues]( https://status.canonical.com/#/incident/KNms6QK9ewuzz-7xUsPsNylV20jEt5kyKsd8A-3ptQHw8b9cLkaHwpIjq0vhMLFLrxbdNiPozwnbJBsW-asYAA== )

PR #408, opened 2026-05-18
refactor(citestwheel): switch base image from devel to base

gforsyth:
New PR because I rebased and force-pushed -- rebuilding `ci-imgs` against cuda `base`

Depends on https://github.com/rapidsai/build-planning/issues/279.

Testing is tracked in https://github.com/rapidsai/build-planning/issues/143#issuecomment-4894101646

As of 2026-07-21 -- all libraries that load shared objects from wheel installs are tested against the cuda-base image, and all fixes to ensure that they _correctly_ load those shared objects have landed in their respective projects.


bdice:
After we merge this, let's work on https://github.com/rapidsai/docs/issues/673 and use `base` as the recommended image flavor in our docs.

jameslamb (approved):
awesome!!!

The images look like they're about 10x smaller ([400ish MB instead of 4ish GB](https://hub.docker.com/r/rapidsai/staging/tags?name=citestwheel-408)) and I saw container spin-up times around 30 seconds on the testing PRs (usually like 3-4 minutes today).

Let's merge this. The testing looks pretty thorough, I'm around to help with whatever other things may break because of this.

vyasr:
That's an awesome speedup. Glad we're finally in a position to do this!

gforsyth:
Let's goooooooooooooo

PR #409, opened 2026-05-18
fix: swap version update/downgrade and general update

gforsyth:
Debugging failures that showed up in https://github.com/rapidsai/ci-imgs/actions/runs/26057524971/job/76609236506?pr=408#step:9:6848 while working on #408 

@jameslamb is seeing the same thing over in `rapidsai/docker` (tracked at https://github.com/rapidsai/docker/issues/878)
 
**update 2026-05-19**

@jameslamb figured out that force-reinstalling `conda` before performing any other `conda` operations fixes the notices loading issue.
I've applied those same changed here and reverted the previous change.

### DIFF
--- a/ci-conda.Dockerfile
+++ b/ci-conda.Dockerfile
@@ -110,8 +110,8 @@ if [[ "$PYTHON_VERSION_PADDED" > "3.12" ]]; then
 else
     PYTHON_ABI_TAG="cpython"
 fi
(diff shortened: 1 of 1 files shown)


KyleFromNVIDIA:
Why does the package need to be reinstalled? Is something corrupting it?

gforsyth:
Yeah -- James wrote up a reproducer and reported it upstream: https://github.com/conda/conda/issues/16126

KyleFromNVIDIA (approved):

jameslamb (approved):
great let's do it, thanks!

PR #417, opened 2026-06-09
Add CUDA 13.3 CI images

bdice:
Now that CUDA 13.3 is available, add CUDA 13.3 CI images.

<hr>

Refs:

* https://github.com/rapidsai/build-planning/issues/286


latest.yaml
```diff
@@ -2,7 +2,7 @@
 # SPDX-License-Identifier: Apache-2.0
 # Define the values used for the "latest" tag
 ci-conda:
-  CUDA_VER: "13.1.1"
+  CUDA_VER: "13.3.0"
```
bdice: This wasn't updated after the CUDA 13.2 migration. We will switch it to 13.3 after the 13.3 migration.
```suggestion
  CUDA_VER: "13.2.0"
```

latest.yaml
```diff
@@ -2,15 +2,15 @@
...
   PYTHON_VER: "3.14"
   LINUX_VER: "ubuntu24.04"
 ci-wheel:
-  CUDA_VER: "13.0.2"
+  CUDA_VER: "13.0.3"
```
gforsyth, line 9:
```suggestion
  CUDA_VER: "13.3.0"
```

Shouldn't this also be `13.3.0`?

bdice: `ci-wheel` is pinned to 13.0.x for (painful) reasons described in https://github.com/rapidsai/build-planning/issues/257 and https://github.com/rapidsai/build-planning/issues/268.

Also the conda images pinned in `latest.yaml` should be `13.2.1` and not `13.3.0` because we haven't completed the 13.3 migration yet. We only roll the `latest.yaml` forward once the migration is complete (this was missed in the 13.2 migration).

matrix.yaml
```diff
@@ -2,10 +2,10 @@
 # SPDX-License-Identifier: Apache-2.0
 CUDA_VER:
   - "12.2.2"
-  - "12.9.1"
-  - "13.0.2"
-  - "13.1.1"
-  - "13.2.0"
+  - "12.9.2"
+  - "13.0.3"
```
jameslamb, line 6: I *think* it could be ok to take these minor pins, let's do this.

But we shoudn't move up our floors on `nvidia-nvjitlink` in wheels.

`torch` CUDA 12.9 wheels are pinned to `cuda-toolkit[nvjitlink]==12.9.1`, so a requirement like `nvidia-nvjitlink>=12.9.2` would be a problem for compatibility. Hopefully using system-provided libnvJitLink from the `12.9.2` images and a runtime requirement like `nvidia-nvjitlink-cu12>=12.9` ([like we have throughout RAPIDS](https://github.com/rapidsai/cudf/blob/8838cf2b21dce89d38c159abcd0023bac7a59360/dependencies.yaml#L782)) should continue to be fine.

jameslamb (approved):

PR #419, opened 2026-06-10
Add zstd support for pyenv Python builds

bdice:
## Summary
- install zstd development headers for Ubuntu pyenv-based images
- build zstd 1.5.7 from source on Rocky, where the distro package is too old for Python 3.14 stdlib zstd support
- apply the same support to both ci-wheel and citestwheel pyenv image families

## Context
conda-package-handling 2.5.0 / conda-package-streaming 0.13.0 expects Python 3.14 to provide compression.zstd or to have backports.zstd installed. The current pyenv-built Python 3.14 in the wheel image lacks _zstd, and backports.zstd is not available for Python 3.14, so unpacking .conda artifacts fails before CI checks run.

## Local validation
- built ci-wheel:zstd-py314-rocky with Python 3.14 on Rocky Linux 8
- verified import compression.zstd succeeds in that image
- verified conda-package-handling 2.5.0 / conda-package-streaming 0.13.0 use compression.zstd as the streaming backend
- built citestwheel:zstd-py314-ubuntu with Python 3.14 on Ubuntu 24.04
- verified import compression.zstd succeeds in that image

### DIFF
--- a/ci-wheel.Dockerfile
+++ b/ci-wheel.Dockerfile
@@ -90,6 +90,7 @@ case "${LINUX_VER}" in
       libsqlite3-dev
       libssl-dev
       libtool
+      libzstd-dev
       openssh-client
       patch
       protobuf-compiler
@@ -193,6 +194,19 @@ case "${LINUX_VER}" in
     make install
     popd
     rm -rf /tmp/openssl*
+    # Python 3.14 adds stdlib compression.zstd and requires libzstd >=1.4.5.
+    # Rocky 8 packages libzstd 1.4.4, so provide a newer zstd before pyenv
+    # builds Python. See https://github.com/pyenv/pyenv/wiki.
+    ZSTD_VERSION=1.5.7
+    pushd /tmp
+    rapids-retry wget -q "https://github.com/facebook/zstd/releases/download/v${ZSTD_VERSION}/zstd-${ZSTD_VERSION}.tar.gz"
+    tar -xzvf "zstd-${ZSTD_VERSION}.tar.gz"
+    cd "zstd-${ZSTD_VERSION}"
+    make -j"$(nproc)" lib-release
+    make -C lib install PREFIX=/usr LIBDIR=/usr/lib64
+    ldconfig
+    popd
+    rm -rf /tmp/zstd*
     ;;
   *)
     echo "Unsupported LINUX_VER: ${LINUX_VER}"
--- a/citestwheel.Dockerfile
+++ b/citestwheel.Dockerfile
@@ -75,6 +75,7 @@ case "${LINUX_VER}" in
       libssl-dev
       libxml2-dev
       libxmlsec1-dev
+      libzstd-dev
       llvm
       make
       patch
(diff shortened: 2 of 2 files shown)


ci-wheel.Dockerfile
```diff
@@ -193,6 +194,18 @@ case "${LINUX_VER}" in
     make install
     popd
     rm -rf /tmp/openssl*
+    # Rocky 8 packages libzstd 1.4.4, but Python 3.14 requires >=1.4.5
```
jameslamb: ugh haha

ci-wheel.Dockerfile
```diff
@@ -193,6 +194,18 @@ case "${LINUX_VER}" in
...
+    # to build the stdlib compression.zstd extension.
+    ZSTD_VERSION=1.5.7
+    pushd tmp
+    rapids-retry wget -q "https://github.com/facebook/zstd/releases/download/v${ZSTD_VERSION}/zstd-${ZSTD_VERSION}.tar.gz"
+    tar -xzvf "zstd-${ZSTD_VERSION}.tar.gz"
+    cd "zstd-${ZSTD_VERSION}"
+    make
+    make install PREFIX=/usr LIBDIR=/usr/lib64
```
jameslamb:
```suggestion
    make -j"$(nproc)" lib-release
    make install-lib PREFIX=/usr LIBDIR=/usr/lib64
```

I think CPython should only need the runtime libraries. We could save some build time and image size by skipping the CLI, which otherwise is pulled in in the `default` target: https://github.com/facebook/zstd/blob/5233c58e6ca0b1c4c6b353ad79649191ed195bdc/Makefile#L45

Let's also take advantage of the multi-CPU runners this builds on to parallelize the build.

ci-wheel.Dockerfile
```diff
@@ -193,6 +194,18 @@ case "${LINUX_VER}" in
     make install
     popd
     rm -rf /tmp/openssl*
+    # Rocky 8 packages libzstd 1.4.4, but Python 3.14 requires >=1.4.5
+    # to build the stdlib compression.zstd extension.
+    ZSTD_VERSION=1.5.7
+    pushd tmp
```
jameslamb:
```suggestion
    pushd /tmp
```

The script later `popd`'s out and does `rm -rf /tmp/zstd*`, but if the working directory wasn't `/` when this `pushd` ran then that's not the right place to look. Let's make this explicit.

jameslamb (approved):
Sounds good, matches my understanding of what's needed from https://github.com/pyenv/pyenv/issues/3296 

Left some small suggestions for improvement, do what you want with them.

bdice:
Addressed the review suggestions:

- changed zstd build to `make -j"$(nproc)" lib-release`
- changed zstd directory handling to `pushd /tmp`
- used `make -C lib install PREFIX=/usr LIBDIR=/usr/lib64` for the library-only install, since zstd 1.5.7 does not have a root-level `install-lib` target

I smoke-tested the corrected commands in `nvidia/cuda:13.0.3-devel-rockylinux8`; they install `zstd.h`, `libzstd.pc`, and `libzstd.so.1.5.7`, and `pkg-config --modversion libzstd` reports `1.5.7`.

Issue #420, opened 2026-06-11
ci-wheel: GCC toolset not being found in CMake builds


jameslamb:
## Description

Wheel builds have started failing recently with errors like this:

```text
  -- The CXX compiler identification is GNU 8.5.0
  -- The CUDA compiler identification is NVIDIA 12.9.86 with host compiler GNU 14.2.1
  -- Detecting CXX compiler ABI info
  -- Detecting CXX compiler ABI info - done
  -- Check for working CXX compiler: /usr/bin/g++ - skipped
  -- Detecting CXX compile features
  -- Detecting CXX compile features - done
  -- Detecting CUDA compiler ABI info
  -- Detecting CUDA compiler ABI info - done
  -- Check for working CUDA compiler: /usr/local/cuda/bin/nvcc - skipped
  -- Detecting CUDA compile features
  -- Detecting CUDA compile features - done
  -- Project libcugraph-python is building for CUDA architectures:
    70-real
    75-real
    80-real
    86-real
    90a-real
    100f-real
    120a-real
    120
...

  -- The C compiler identification is GNU 14.2.1
  -- Detecting C compiler ABI info
  -- Detecting C compiler ABI info - done
  -- Check for working C compiler: /opt/rh/gcc-toolset-14/root/usr/bin/gcc - skipped
  -- Detecting C compile features
  -- Detecting C compile features - done
  CMake Error at /__w/cugraph/cugraph/cpp/CMakeLists.txt:24 (message):
    GCC compiler must be at least 9.3


  -- Configuring incomplete, errors occurred!

  *** CMake configuration failed
  error: subprocess-exited-with-error
```

([cugraph example](https://github.com/rapidsai/cugraph/actions/runs/27367676475/job/80871782107?pr=5555#step:13:231))

We'd expect to see GCC / G++ 14:

https://github.com/rapidsai/ci-imgs/blob/2bbcab77116642b15f764df2b33fe3ff1bd57b90/ci-wheel.Dockerfile#L143-L144

https://github.com/rapidsai/ci-imgs/blob/2bbcab77116642b15f764df2b33fe3ff1bd57b90/ci-wheel.Dockerfile#L184-L187

## Reproducible Example

N/A

## Notes

N/A

jameslamb:
To start... this all looks correct.

```shell
docker run \
  --rm \
  --pull always \
  -it rapidsai/ci-wheel:26.08-cuda12.9.2-rockylinux8-py3.14 \
  bash
```

[...]

PR #421, opened 2026-06-11
ci-wheel: use GCC toolset when building CPython and its dependencies

jameslamb:
Contributes to #420 (hopefully "fixes", but we'll see 😅)

As described there, we observed wheel builds using `scikit-build-core` falling back to the GCC 8.5 at `/usr/bin/g++`.

```text
  -- The CXX compiler identification is GNU 8.5.0
  -- The CUDA compiler identification is NVIDIA 12.9.86 with host compiler GNU 14.2.1
  -- Detecting CXX compiler ABI info
  -- Detecting CXX compiler ABI info - done
  -- Check for working CXX compiler: /usr/bin/g++ - skipped
```

Despite enabling the GCC 14 toolset:

https://github.com/rapidsai/ci-imgs/blob/2bbcab77116642b15f764df2b33fe3ff1bd57b90/ci-wheel.Dockerfile#L143-L144

It looks like the root cause is something like:

* `sysconfig.get_config_var("CXX")` returns the compiler the Python interpreter was built with
* at some point recently (possibly https://github.com/python/cpython/pull/148298), CPython started recording an absolute path like `/usr/bin/g++`
* we build CPython from source in the `ci-wheel` images used in RAPIDS, `/usr/bin/g++` is the first C++ compiler on `PATH` at the time that build happens
* `scikit-build-core` reads that value when configuring CMake projects ([scikit-build/scikit-build-core - src/scikit_build_Core/builder/generator.py](https://github.com/scikit-build/scikit-build-core/blob/05948f43ffa21f5ab84c0c2d239f1509d5e39882/src/scikit_build_core/builder/generator.py#L118))
* activating the GCC toolset doesn't work because that absolute path circumvents any `PATH` lookup at CMake configure time... `/usr/bin/g++` is used unconditionally

This attempts to fix it by activating the GCC toolset before building CPython.

## Notes for Reviewers

I haven't tested this locally, getting the PR up to start CI before I have to step away for a bit. Will try to return to it tonight.

I also think I have an idea of which CPython change caused this, will report an issue upstream tonight.

### DIFF
--- a/ci-wheel.Dockerfile
+++ b/ci-wheel.Dockerfile
@@ -237,7 +237,16 @@ case "${LINUX_VER}" in
     pyenv install --verbose "${RAPIDS_PY_VERSION}"
     ;;
   "rockylinux"*)
-    # Need to specify the openssl location because of the install from source
+    # Activate gcc-toolset so its toolchain is used when building CPython and other libraries
+    # from source below.
+    #
+    # In some situations, CPython's ./configure may record an absolute path for
+    # the compiler. Not ideal, but if that's going to happen we want it to be the one
+    # used for building RAPIDS packages, because scikit-build-core retries that value
+    # with `sysconfig.get_config_var("CXX")`.
+    source /etc/profile.d/enable_devtools.sh
+
+    # Need to specify the openssl location because of the install from source.
     CPPFLAGS="-I/usr/include/openssl" LDFLAGS="-L/usr/lib" pyenv install --verbose "${RAPIDS_PY_VERSION}"
     ;;
   *)


vyasr (approved):
Good sleuthing! I'll trust you to go forth and test/verify, but the reasoning sounds plausible and the links you posted pass the smell test.

bdice (approved):
Validated with:

```bash
docker run --rm rapidsai/staging:ci-wheel-421-26.08-cuda12.9.2-rockylinux8-py3.14 \
  python -c 'import sysconfig; print("CC =", sysconfig.get_config_var("CC")); print("CXX =", sysconfig.get_config_var("CXX"))'
```

Output:
```bash
CC = gcc -pthread
CXX = /opt/rh/gcc-toolset-14/root/usr/bin/g++ -pthread
```

PR #423, opened 2026-06-22
feat(renovate): add 7-day cooldown before any updates

gforsyth:
Part of rapidsai/build-planning#273

Adds a 7-day cooldown period before renovate proposes any updates

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -8,6 +8,13 @@
   ],
   "packageRules": [
     {
+      "description": "Wait for 7 days post-release before updating",
+      "matchPackageNames": [
+        "*"
+      ],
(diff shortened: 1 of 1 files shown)


jameslamb (approved):

jameslamb:
1. IMO this is fine to admin-merge without a rebuild
2. we should set up `changed-files` here so similar changes don't require rebuilds in the future. This project's CI is particularly expensive, skipping it for skippable changesets could be nice.

gforsyth:
> * 
> 
> * we should set up `changed-files` here so similar changes don't require rebuilds in the future. This project's CI is particularly expensive, skipping it for skippable changesets could be nice.

yeah, that's a good call

PR #426, opened 2026-06-25
feat: add `changed-files` CI filter to `pr.yaml` to skip builds for certain changesets

gforsyth:
As proposed by @jameslamb in #423, we don't need to run the (quite expensive) CI here when we do things like update `renovate.json`

### DIFF
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -58,4 +58,5 @@ jobs:
           - '!LICENSE'
           - '!README.md'
           - '!SECURITY.md'
+          - '!ci/release/update-version.sh
           - '!renovate.json'
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -20,6 +20,7 @@ jobs:
   pr-builder:
     needs:
       - build-images
+      - changed-files
     uses: rapidsai/shared-workflows/.github/workflows/pr-builder.yaml@main
     if: always()
     with:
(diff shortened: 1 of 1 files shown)


.github/workflows/pr.yaml
```diff
@@ -30,8 +31,31 @@ jobs:
...
+          - '!.github/CODEOWNERS'
+          - '!.github/copy-pr-bot.yaml'
+          - '!.github/ops-bot.yaml'
+          - '!.github/zizmor.yml'
+          - '!.gitignore'
+          - '!.pre-commit-config.yaml'
+          - '!CONTRIBUTING.md'
+          - '!LICENSE'
+          - '!README.md'
+          - '!SECURITY.md'
```
jameslamb, line 52:
```suggestion
          - '!SECURITY.md'
          - '!ci/release/update-version.sh
```

`update-version.sh` doesn't affect the built images either, let's add it.

otherwise, agree with this list

.github/workflows/pr.yaml
```diff
@@ -30,8 +31,31 @@ jobs:
       pull-requests: read
   build-images:
     uses: ./.github/workflows/build-and-publish-images.yaml
+    needs: changed-files
     with:
       build_type: pull-request
     secrets:
       GPUCIBOT_DOCKERHUB_USER: ${{ secrets.GPUCIBOT_DOCKERHUB_USER }}
       GPUCIBOT_DOCKERHUB_TOKEN: ${{ secrets.GPUCIBOT_DOCKERHUB_TOKEN }}
+  changed-files:
```
jameslamb, line 32: Could we put this above `build-images`?

I think it's generally preferable for things referenced in `needs:` to occur earlier in the file, so the order from top to bottom roughly matches the order of execution.

jameslamb (approved):
Love this, thank you!

2 very minor comments, do what you want with them, I don't need to re-review.

.github/workflows/pr.yaml
gforsyth: Will do, then I'll admin merge

gforsyth:
admin-merging to save on CI here, this has just run and the changes made won't impact builds

PR #444, opened 2026-07-21
Add gdb to citestwheel images

wence-:
For some test runs in rapidsmpf and cudf-polars, we use gdb to attach to, and capture stack traces of, hanging tests. This helps diagnosing CI errors that are hard to reproduce locally. This requires gdb in the citestwheel image because, unlike conda images, we have no way of installing it locally.

### DIFF
--- a/citestwheel.Dockerfile
+++ b/citestwheel.Dockerfile
@@ -62,6 +62,7 @@ case "${LINUX_VER}" in
       build-essential
       ca-certificates
       curl
+      gdb
       git
       jq
       libbz2-dev
@@ -111,6 +112,7 @@ case "${LINUX_VER}" in
       curl
       dnf-plugins-core
       gcc
+      gdb
       git
       jq
       libffi-devel


jameslamb (approved):
> _unlike conda images, we have no way of installing it locally_

I didn't understand this point, as it seems to me `apt-get install gdb` in a `citestwheel` container locally should work just as `conda install gdb` in a `ci-conda` one does. For my own understanding, could you clarify that?

But either way, fine to include this.

wence-:
> > _unlike conda images, we have no way of installing it locally_
> 
> I didn't understand this point, as it seems to me `apt-get install gdb` in a `citestwheel` container locally should work just as `conda install gdb` in a `ci-conda` one does. For my own understanding, could you clarify that?
> 
> But either way, fine to include this.

needs sudo, no? and I am not sure if that's available. Also, pulling the entire apt repository and solving every CI run seems bad.

jameslamb:
> _needs sudo, no?_

Nope.

```shell
docker run \
  --rm \
  -it rapidsai/citestwheel:26.08-latest \
  bash

apt-get update
apt-get install -y gdb

gdb --version
# GNU gdb (Ubuntu 17.1-2ubuntu1) 17.1
```

> _pulling the entire apt repository and solving every CI run seems bad_

🤷🏻 that's conceptually what we're also doing with `conda install gdb`.

The tradeoff of pre-installing is that CI jobs everywhere in RAPIDS (including the many that don't use `gdb`) pay the download and unpacking cost of everything we add to these images.

It's totally fine to include, `gdb` is generically useful for these jobs. I really did mean it when I approved this.

vyasr:
Just to close the loop on the permissions side of the question. we never switch users in the CI containers so they're always running as root.

